PR #2, opened 2016-08-10
Refactor into packages

kvigen:
Instead of having one giant main, we should probably have multiple packages, each of which is a reasonable size. This will make it easier to work on multiple parts of the code as once, and also make it easier to understand what's going on and change.

This change puts the models, client, and server code generation in separate packages, and also adds a Swagger package for common operations. It doesn't have any changes in functionality.

The code could still use a lot of work, but I think this is a step in the right direction and will let us make the changes we need.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,7 +7,7 @@ $(eval $(call golang-version-check,1.6))
 
 test:
 	rm -rf generated/*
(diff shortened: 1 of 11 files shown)


rgarcia:
lgtm!

PR #3, opened 2016-08-10
Basic drone yml

kvigen:
Will add publishing in the future


rgarcia:
lgtm, will have to think about how we want to publish this (leaning towards docker image)

PR #4, opened 2016-08-10
Run the tests

kvigen:
Actually run the tests I added in `make test`. Not sure if this is the way we want to test long-term, but at least we're doing more than just making sure it compiles.


rgarcia:
oops, let this one slip through the cracks. LGTM!

PR #6, opened 2016-08-11
build and publish a docker image with wag binary

rgarcia:
Usage is slightly awkward, but I think a docker image will be most convenient while we're making changes frequently:

`docker run -v $(GOPATH)/src:/gopath/src -w /gopath/src/$(PKG) -i -t clever/wag:<commit> -file swagger.yml -package $(PKG)/generated`

Next up is a PR on dev-handbook with a swagger.mk that runs this command ^

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -11,5 +11,18 @@ notify:
     on_started: false
     on_success: false
     webhook_url: $$slack_webhook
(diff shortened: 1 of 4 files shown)


kvigen:
lgtm

PR #8, opened 2016-08-11
Graceful shutdown

kvigen:
This change adds graceful shutdown to the generated code. It uses a fairly popular graceful shutdown library.

It changes the interface for setting up a server a bit. I think it's generally good since it gives us a bit more flexibility for how we start up the server, and that's something consumers are unlikely to want to change. And, if they do want to change it, this change doesn't preclude us from adding options later.


kvigen:
Discussed with Infra and these seems to satisfy their current needs.

generated/server/router.go
```diff
@@ -1,18 +1,37 @@
...
+
+	"gopkg.in/tylerb/graceful.v1"
 )
 
 type contextKey struct{}
 
-func SetupServer(r *mux.Router, c Controller) http.Handler {
+type Server struct {
+	Handler http.Handler
+	Port    int
```
rgarcia: should we leave these unexported for now?

rgarcia:
lgtm!

generated/server/router.go
kvigen: the reason i made `Handler` public is that I needed it for my test: `httptest.NewServer(s.Handler)`. I imagine others are going to want to do something similar. Maybe I should expose a `ServeForTesting` method instead? Any other ideas?

will make port private

rgarcia: public `Handler` seems fine

PR #9, opened 2016-08-12
use go-bindata to package hardcoded assets

rgarcia:
Also changes the hardcoded go files to be prepended with underscores so that the compiler ignores them.

`go-bindata` turns a directory into a .go file with files available via an `Asset` function.


kvigen:
lgtm!

PR #10, opened 2016-08-12
Un-hardcode the generated/ path

rgarcia:
This changes the `Generator` object to only write files relative to the
package name passed in as an input.

It also removes the old `Generator` object from main.go. Since there was
some direct access to the `bytes.Buffer` contained in this `Generator`,
I added a passthrough `Write` method so that the `Generator` can be used as an `io.Writer`

### DIFF
--- a/swagger/swagger.go
+++ b/swagger/swagger.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"go/format"
 	"io/ioutil"
+	"os"
 	"sort"
 	"strings"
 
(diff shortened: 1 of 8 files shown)


swagger/swagger.go
```diff
@@ -5,29 +5,38 @@ import (
...
-	buf bytes.Buffer
+	PackageName string
+	buf         bytes.Buffer
 }
 
 func (g *Generator) Printf(format string, args ...interface{}) {
 	fmt.Fprintf(&g.buf, format, args...)
 }
 
+func (g *Generator) Write(p []byte) (n int, err error) {
```
kvigen: 👍

kvigen:
lgtm, this is a nice improvement!

PR #12, opened 2016-08-15
change port to an address string

rgarcia:
This seems more in line with what `Listen` functions take in, and also avoids having to do int -> string conversions (before this change we were doing it incorrectly).

### DIFF
--- a/generated/server/router.go
+++ b/generated/server/router.go
@@ -1,7 +1,6 @@
 package server
 
 import (
-	"errors"
 	"net/http"
 	"time"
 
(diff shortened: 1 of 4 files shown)


generated/server/router.go
```diff
@@ -17,18 +16,15 @@ type contextKey struct{}
...
-	port    int
+	addr    string
 }
 
 func (s Server) Serve() error {
 	// Give the sever 30 seconds to shut down
-	graceful.Run(":"+string(s.port), 30*time.Second, s.Handler)
-
-	// This should never return
-	return errors.New("This should never happen")
```
rgarcia: without this change I was seeing the server crash w/o an error message

kvigen:
lol, thanks for fixing, lgtm!

PR #13, opened 2016-08-16
Add required and validation checks for input parameters

kvigen:
This change updates the way we handle input parameters in a few ways:
1. It makes non-required fields pointers
2. It errors if a field is required and isn't specified
3. It validates the input parameters based on the swagger spec. Since we don't support array or schema type input parameters, this isn't too difficult

We considered using go-swagger's internal codegen for this, but instead decided it was easier to just use their validation libraries and generate the code ourselves.


README.md
```diff
@@ -140,11 +140,6 @@ make test
 ## Future Features
 
 These are the features we're planning on building in the near future:
-- Stricter input validation
-  - Support for the required field in the Swagger spec
-  - At least some of the validation defined in http://swagger.io/specification/#parameterIn. In particular we want to be able to support checking whether an input parameter is a valid object ID.
-- More robust server defaults
-  - SIGTERM handling
```
rgarcia: seeing this diff makes me think that it would be useful to have a list of features at the top of this readme

client/genclients.go
```diff
@@ -176,16 +176,19 @@ func successResponse(outputName string) string {
 }
 
 func convertParamToString(p spec.Parameter) string {
-	capParamName := swagger.Capitalize(p.Name)
+	valToSet := fmt.Sprintf("i.%s", swagger.Capitalize(p.Name))
+	if !p.Required {
+		valToSet = "*" + valToSet
+	}
 	switch p.Type {
```
rgarcia: might be useful in these switch statements to link to the swagger spec that enumerates the different values

generated/models/inputs.go
```diff
@@ -2,34 +2,77 @@ package models
...
+		return err
+	}
+
+	if err := validate.Minimum("maxPages", "query", *i.MaxPages, -5.000000, false); err != nil {
+		return err
+	}
+
+	if err := validate.MultipleOf("maxPages", "query", *i.MaxPages, 0.500000); err != nil {
+		return err
+	}
```
rgarcia: nice

generated/server/handlers.go
```diff
@@ -57,19 +58,24 @@ func GetBooksHandler(ctx context.Context, w http.ResponseWriter, r *http.Request
...
+	authorTmp := authorStr
+	input.Author = &authorTmp
+
 	availableStr := r.URL.Query().Get("available")
-	var err error
-	input.Available, err = strconv.ParseBool(availableStr)
-	if err != nil {
-		return nil, err
-	}
+	availableTmp, err := strconv.ParseBool(availableStr)
```
rgarcia: looks like this err is never checked

generated/server/handlers.go
```diff
@@ -158,9 +172,10 @@ func CreateBookHandler(ctx context.Context, w http.ResponseWriter, r *http.Reque
 func NewCreateBookInput(r *http.Request) (*models.CreateBookInput, error) {
 	var input models.CreateBookInput
 
-	if err := json.NewDecoder(r.Body).Decode(&input.NewBook); err != nil {
-		return nil, err
-	}
+	var err error
+	_ = err
+
+	err = json.NewDecoder(r.Body).Decode(input.NewBook)
```
rgarcia: nit: this code now looks much more auto-generated (I like the previous `if err := json.NewDecoder(...` better)

generated/server/handlers.go
```diff
@@ -57,19 +58,24 @@ func GetBooksHandler(ctx context.Context, w http.ResponseWriter, r *http.Request
...
+	stateStr := r.URL.Query().Get("state")
+	stateTmp := stateStr
+	input.State = &stateTmp
+
 	maxPagesStr := r.URL.Query().Get("maxPages")
-	input.MaxPages, err = strconv.ParseFloat(maxPagesStr, 64)
-	if err != nil {
-		return nil, err
-	}
+	maxPagesTmp, err := strconv.ParseFloat(maxPagesStr, 64)
```
rgarcia: same here

rgarcia:
I think I've mentioned this but one other swagger feature that I sort of lump together with input validation is their "data type formats": http://swagger.io/specification/#dataTypeFormat Most useful of which is `date-time`. It'd be great to add support for this in this PR, but if it's too much can you add it as a future feature?

generated/server/handlers.go
kvigen: when i originally wrote it i thought it made sense to ignore errors if the parameter wasn't required, but now that I think about it, i should error here if they did specify something and it failed

kvigen:
derp, i meant to add the "data type formats" here and just forgot to do it. I think it should be pretty straightforward. Just a new parse / validate function in the switch statements:
`formats.Parse("date", raw)`
`validate.FormatOf("district_id", "query", "date", string(o.DistrictID), formats)`

I will update with all 5 that are missing.

README.md
kvigen: agreed, will add this in a future PR

generated/server/handlers.go
kvigen: changed it back

kvigen:
@rgarcia I added the formats, except I left off binary for now since it converted to an `io.Reader` / `io.Writer` and I didn't really want to deal with it

I also refactored the code quite a bit, so this because a bit bigger change than I originally anticipated. I put all the parameter validation logic into a separate file, which I think makes it a bit clearer what we support, and how there are 4 common operations on the types. I also added quite a few `if x != nil` checks where I probably should have added them before. If this PR has too much going on, let me know and I can try to break it up.

There are still a few things I could do to clean up, but I figured this was enough for one PR.

generated/models/inputs.go
```diff
@@ -2,34 +2,116 @@ package models
...
+	Published   *strfmt.Date
+	Completed   *strfmt.DateTime
+	MaxPages    *float64
+	MinPages    *int32
+	PagesToTime *float32
 }
 
 func (i GetBooksInput) Validate() error {
+	if i.State != nil {
+
```
rgarcia: nit: blank line

swagger/parameter.go
```diff
@@ -0,0 +1,238 @@
...
+
+// accessString returns a string with the access of a variable in the struct named 'i'. For example:
+// *i.Length
+func accessString(param spec.Parameter) string {
+	pointer := ""
+	if !param.Required {
+		pointer = "*"
+	}
+	return fmt.Sprintf("%si.%s", pointer, Capitalize(param.Name))
+}
```
rgarcia: :clap: I like splitting this out

rgarcia:
lgtm!

PR #14, opened 2016-08-16
Update the way we use contexts

kvigen:
This change updates the way we use contexts in our Go server and client code now that Go 1.7 has better support for them. This change does a few things:

1. Upgrades this repo to go 1.7
2. Gets rid of the clunky middleware handling in the server that adds removes things from contexts
3. Changes the way we handle contexts in the client slightly
4. Adds a bunch of tests related to contexts, one of which exposed a bug.


test/server_test.go
```diff
@@ -68,3 +72,37 @@ func TestClientSideError(t *testing.T) {
 	_, ok := err.(models.DefaultInternalError)
 	assert.True(t, ok)
 }
+
+type MiddlewareContextTest struct {
```
kvigen: at some point we should probably use gomock, or something like that for this. if nothing else, for an example that others can copy

generated/server/middleware.go
```diff
@@ -6,71 +6,21 @@ import (
 	"gopkg.in/Clever/kayvee-go.v3/logger"
 	kvMiddleware "gopkg.in/Clever/kayvee-go.v3/middleware"
 
-	"golang.org/x/net/context"
-
-	gContext "github.com/gorilla/context"
```
rgarcia: :boom:

generated/server/middleware.go
```diff
@@ -6,71 +6,21 @@ import (
...
-	c(ctx, w, r)
-}
-
-// modifyContextExample is sample middleware that modifies the context. We can remove it when
-// we have real middleware that uses the context.
-func modifyContextExample(c ContextHandler) ContextHandler {
-	return ContextHandlerFunc(func(ctx context.Context, w http.ResponseWriter, r *http.Request) {
-		ctx = context.WithValue(ctx, "addedKey", "addedValue")
-		c.ServeHTTPContext(ctx, w, r)
-	})
```
rgarcia: go 1.7 for the win

hardcoded/_doer.go
```diff
@@ -3,34 +3,35 @@ package client
 import (
 	"fmt"
 	"net/http"
+
+	"golang.org/x/net/context/ctxhttp"
 )
 import "golang.org/x/net/context"
 
 import opentracing "github.com/opentracing/opentracing-go"
```
rgarcia: nit: these imports should go into the above `import ( ... )` block

test/server_test.go
rgarcia: 👍

rgarcia:
lgtm!

PR #15, opened 2016-08-16
Remove generated controller file

kvigen:
We don't expect people to actually use this file and it just clutters
things. In the process I also removed the impl/ directory since it was
used for making sure the generated code built, and we have tests to
verify that (and more) now.


rgarcia:
lgtm

PR #16, opened 2016-08-18
Factor out validation logic

kvigen:
This change factors out the validation logic and updates the Makefile / tests a bit to make sure all the tests, formatting, vetting, etc.. are run.

I also added a check to ensure that path parameters are required.


Makefile
```diff
@@ -1,6 +1,6 @@
 include golang.mk
 .DEFAULT_GOAL := test # override default goal set in library makefile
-.PHONY: test build
+.PHONY: test build hardcoded.go
```
kvigen: wasn't totally sure the best way to setup the makefile, so let me know if you have a better idea

Makefile
```diff
@@ -9,17 +9,24 @@ build: hardcoded.go
...
+$(PKGS): golang-test-all-deps
+	$(call golang-test-all,$@)
 
 $(GOPATH)/bin/go-bindata:
 	go get -u github.com/jteeuwen/go-bindata/...
 
 hardcoded.go: $(GOPATH)/bin/go-bindata hardcoded/*
 	$(GOPATH)/bin/go-bindata -o hardcoded.go hardcoded/
+	# gofmt doesn't like what go-bindata creates
+	gofmt hardcoded.go > temp
```
rgarcia: I think there's a `gofmt` flag that does the edit in place, which would eliminate the need for this temp file

Makefile
rgarcia: Making the `hardcoded.go` target PHONY means that it will be regenerated even if none of its dependencies have changed. Right now the main dependencies are the contents of the hardcoded directory (`hardcoded/*` below), which seems sufficient to keep the target non-PHONY? Were you seeing `make` not update this file when changed were made to this directory?

validation/validation.go
```diff
@@ -13,44 +13,45 @@ var alphaNumRegex = regexp.MustCompile("^[a-zA-Z][a-zA-Z0-9]*$")
...
-		fmt.Errorf("Produces not supported in swagger operations")
+		return fmt.Errorf("Produces not supported in swagger operations")
 	}
 	if len(s.Schemes) != 0 {
-		fmt.Errorf("Schemes not supported in swagger operations")
+		return fmt.Errorf("Schemes not supported in swagger operations")
 	}
 	if len(s.Security) != 0 {
-		fmt.Errorf("Security not supported in swagger operations")
+		return fmt.Errorf("Security not supported in swagger operations")
```
rgarcia: ha, oops 😄

rgarcia:
lgtm

Makefile
kvigen: ah nice

validation/validation.go
kvigen: thank you `go vet` :)

Makefile
kvigen: makes sense. I think I was confused at the time and was trying random things.

PR #17, opened 2016-08-18
Support for default values

kvigen:
This change adds support for default values from the swagger schema. It converts the default value to a string and sets it as the value if it isn't specified.

In go-swagger adding a default to a input parameter doesn't make it a non-pointer. My intuition is that it should because the client should never see a nil value, but I held off on implemented that until I got buy-in.

### DIFF
--- a/generated/models/inputs.go
+++ b/generated/models/inputs.go
@@ -41,7 +41,7 @@ func (i GetBooksInput) Validate() error {
 		}
 	}
 	if i.MaxPages != nil {
-		if err := validate.Maximum("maxPages", "query", float64(*i.MaxPages), 1.000000, false); err != nil {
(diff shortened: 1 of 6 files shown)


rgarcia:
what about defaults for non-query string parameters?

rgarcia:
lgtm

kvigen:
My thoughts for non-query parameters:
1. I've disallowed non-required path parameters. Seems like path params should always be required.
2. Headers can have defaults. It's a bit weird, but I thought it wouldn't hurt and we get it for free (I added it to the test)
3. Body parameters have to be schemas, so we get whatever defaults go-swagger provides from those.

Let me know if you think we should change of any of this.

PR #18, opened 2016-08-20
generate mocks for the Controller interface

rgarcia:
This adds a `go generate` directive in the controller file, which makes it possible to generate a mock by calling `go generate` on this package. I have a modification to swagger.mk that calls `go generate` and it seems to work...


rgarcia:
this copies a pattern @natebrennand introduced in the pipeline-finalizer/email package

natebrennand:
@rgarcia, can you commit mockgen's generated files?

rgarcia:
@natebrennand good call fb659f6

generated/server/mock_controller.go
```diff
@@ -0,0 +1,64 @@
+// Automatically generated by MockGen. DO NOT EDIT!
+// Source: interface.go
+
+package server
```
kvigen: hmm, would it make sense to move this to the models or client package to avoid having to import the server package? maybe we can re-visit when we have a separate client interface?

kvigen:
lgtm! only question is about where the interface / generated files belong, and we can punt on that

generated/server/mock_controller.go
rgarcia: Yeah, I think we should consider this when we add an interface type in the `client` package. For now I'm fine generating this mock and using it as a mock of the client temporarily

PR #21, opened 2016-08-23
Don't require produces / consumes

kvigen:
Hit this when trying to migrate the timeline service. Didn't seem like we really needed it so I relaxed the constraint.


rgarcia:
lgtm

PR #24, opened 2016-08-24
Support string array input parameters

kvigen:
Adds support for string arrays as input parameters. There are a couple limitations that I'm punting on for a bit so we can unblock timeline.

1. It doesn't escape characters correctly. For example if a parameter has a comma in it, it will treat that as two input parameters. This is a problem with go-swagger and will require us changing what they have a bit
2. It only supports strings. At some point we may want to support arbitrary fields, but supporting things like nested arrays is a bit tricker, so I put it off.

### DIFF
--- a/client/genclients.go
+++ b/client/genclients.go
@@ -47,6 +47,8 @@ func (c Client) WithRetries(retries int) Client {
 
 `)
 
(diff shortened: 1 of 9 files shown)


rgarcia:
lgtm

PR #25, opened 2016-08-25
Use raw model type if only one success response type

kvigen:
This change makes it so that an operation's interface uses the raw type if there's only one success response type. This makes the code easier to deal with (see both the sample server and client code).

To do that I separated out the code common to the interface and types into a separate package. Then I changed that code to be smart about when to use the raw type and when to add an interface. I also changed the gen models code to not generate the interface types when they weren't necessary.

### DIFF
--- a/generated/client/client.go
+++ b/generated/client/client.go
@@ -60,7 +60,7 @@ func JoinByFormat(data []string, format string) string {
 	}
 	return strings.Join(data, sep)
 }
-func (c Client) GetBooks(ctx context.Context, i *models.GetBooksInput) (*[]models.Book, error) {
+func (c Client) GetBooks(ctx context.Context, i *models.GetBooksInput) ([]models.Book, error) {
 	path := c.BasePath + "/v1/books"
 	urlVals := url.Values{}
 	var body []byte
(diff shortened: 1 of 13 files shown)


generated/client/client.go
```diff
@@ -60,7 +60,7 @@ func JoinByFormat(data []string, format string) string {
 	}
 	return strings.Join(data, sep)
 }
-func (c Client) GetBooks(ctx context.Context, i *models.GetBooksInput) (models.GetBooksOutput, error) {
+func (c Client) GetBooks(ctx context.Context, i *models.GetBooksInput) (*[]models.Book, error) {
```
kvigen: I'm not sure whether I like the pointer to the slice. Should it just be `[]models.Book`? Should it be `[]*models.Book`?

rgarcia: yeah I would prefer either `[]models.Book` or `[]*models.Book` over the pointer to a slice

swagger/operation.go
```diff
@@ -0,0 +1,103 @@
...
+		if !strings.HasPrefix(ref, "#/definitions/") {
+			return "", fmt.Errorf("schema.$ref has undefined reference type. Must start with #/definitions")
+		}
+		def := ref[len("#/definitions/"):]
+		if includeModels {
+			def = "models." + def
+		}
+		return "[]" + def, nil
+	}
+}
```
rgarcia: I like these new methods

rgarcia:
lgtm! excited for this change

generated/client/client.go
kvigen: cool, i'll go with `[]models.Book` for now

PR #26, opened 2016-08-25
golang.org/x/net/context -> context

rgarcia:
Change references to use the builtin context package.

Also update godeps


kvigen:
lgtm!

PR #29, opened 2016-08-26
Support no return types

kvigen:
Before we would convert a single, empty response into a string type. This was a bit confusing for implementers, so we decided that if you didn't have any response type it should be left out of the interface.

This made the generation code a bit more complicated, but I think it's probably worth it given that a decent number of endpoints won't have a return type associated with them (health checks and potentially posts).

I want to clean up the templating a bit in the client library since the code is a bit harder to follow than i would like, but I will do that in a follow up change.

### DIFF
--- a/generated/client/client.go
+++ b/generated/client/client.go
@@ -254,3 +254,47 @@ func (c Client) CreateBook(ctx context.Context, i *models.CreateBookInput) (*mod
 
 	}
 }
+
+func (c Client) HealthCheck(ctx context.Context, i *models.HealthCheckInput) error {
+	path := c.BasePath + "/v1/health/check"
+	urlVals := url.Values{}
+	var body []byte
+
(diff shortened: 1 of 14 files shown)


generated/client/client.go
```diff
@@ -251,6 +258,50 @@ func (c Client) CreateBook(ctx context.Context, i *models.CreateBookInput) (*mod
 
 	default:
 		return nil, models.DefaultInternalError{Msg: "Unknown response"}
+	}
+}
+
+func (c Client) HealthCheck(ctx context.Context, i *models.HealthCheckInput) error {
```
rgarcia: looks like `i` is unused, which would be slightly confusing from the perspective of someone consuming this as a client. I think this is an edge case, though.

rgarcia:
lgtm!

PR #30, opened 2016-08-30
Make more variables not required

kvigen:
Did a manual inspection of the code and just found these two.

At some point I'll add a test with a minimal swagger definition so we
can catch these proactively, but for now I just want to unblock Benji.


rgarcia:
lgtm

PR #31, opened 2016-08-30
Write correct status code

kvigen:
Before this change we were writing a 200 for all success responses. This changes this to actually return the right values.

Along the way I also changed the empty response from `string` to `struct{}` since I thought that made the code a bit clearer.


rgarcia:
lgtm

PR #32, opened 2016-08-30
Camel case params

kvigen:
This change converts all the input params to camel case because that's idiomatic in Go.

We didn't just force all consumers to use camel case parameter names because that wouldn't necessarily work in all languages, and being able to support snake case variable names makes migrating some services easier.

### DIFF
--- a/generated/server/handlers.go
+++ b/generated/server/handlers.go
@@ -172,18 +172,18 @@ func NewGetBooksInput(r *http.Request) (*models.GetBooksInput, error) {
 		input.MaxPages = &maxPagesTmp
 
 	}
-	minPagesStr := r.URL.Query().Get("minPages")
-	if len(minPagesStr) == 0 {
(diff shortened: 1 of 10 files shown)


kvigen:
cc @bstein-clever

rgarcia:
@kvigen lgtm besides one consideration about lint

generated/server/handlers.go
```diff
@@ -239,17 +239,17 @@ func NewGetBookByIDInput(r *http.Request) (*models.GetBookByIDInput, error) {
...
+		var book_idTmp int64
+		book_idTmp, err = swag.ConvertInt64(book_idStr)
 		if err != nil {
 			return nil, err
 		}
-		input.BookID = bookIDTmp
+		input.BookID = book_idTmp
 
 	}
 	authorizationStr := r.Header.Get("authorization")
```
kvigen: good point, let me see if i can clean that up

PR #33, opened 2016-08-30
Don't add input parameter argument if not necessary

kvigen:
If we don't need the input parameter argument don't add it to the
interface. It makes both the client and server implementations of these
methods a bit simpler, though methods with no arguments are fairly
unlikely from a client perspective as they usually come up in health
checks. It makes the codegen logic a bit trickier.

Not sure if it's worth it.

### DIFF
--- a/generated/client/client.go
+++ b/generated/client/client.go
@@ -261,7 +261,7 @@ func (c Client) CreateBook(ctx context.Context, i *models.CreateBookInput) (*mod
 	}
 }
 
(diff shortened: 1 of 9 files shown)


kvigen:
cc @rgarcia this is what it would take. It's not too complicated right now, so it might be worth it. Not sure how much more complicated this code would become. Let me know what you think.

rgarcia:
@kvigen I like it--seems simple / not terribly complex

kvigen:
cool, i'll push it then

PR #35, opened 2016-08-31
Fix bug with camel case change

kvigen:
Missed a place we should have been using camel case.

Will add more systematic tests


kvigen:
cc @bstein-clever

rgarcia:
lgtm

PR #36, opened 2016-08-31
Call WriteHeader after changing headers

kvigen:
According to the docs: "Changing the header after a call to
WriteHeader (or Write) has no effect unless the modified headers
were declared as trailers"

So let's do it in the other order.

Will add some tests tomorrow


kvigen:
cc @samfishman

rgarcia:
lgtm, should add tests

PR #39, opened 2016-08-31
simplify interface for single-input operations

rgarcia:
This is a breaking change to the server / client interface. It changes the interface definition for an operation like this

```yaml
    post:
      operationId: createBook
      description: Creates a book
      parameters:
        - name: newBook
          in: body
          schema:
            $ref: "#/definitions/Book"
      responses:
        200:
          description: "Success"
          schema:
            $ref: "#/definitions/Book"
```

from 

```go
CreateBook(ctx context.Context, i *models.CreateBookInput) (*models.Book, error)
...
type CreateBookInput struct {
	NewBook *Book
}
```

to

```go
CreateBook(ctx context.Context, i *models.Book) (*models.Book, error)
```

Generally speaking it will remove the wrapper input struct for any operation that has a single input in the body of the request.

cc @carolinemodic who brought this up in reviewing matchmaker-service.

### DIFF
--- a/client/genclients.go
+++ b/client/genclients.go
@@ -26,6 +26,7 @@ import (
 		"strconv"
 		"encoding/json"
 		"strconv"
+
 		"%s/models"
 )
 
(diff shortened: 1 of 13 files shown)


swagger/operation.go
```diff
@@ -7,13 +7,30 @@ import (
...
 // Interface returns the interface for an operation
 func Interface(op *spec.Operation) string {
 	capOpID := Capitalize(op.ID)
 
-	// Don't add the input parameter argument unless there are some arguments
+	// Don't add the input parameter argument unless there are some arguments.
+	// If a method has a single input parameter, and it's a schema, make the
+	// generated type for that schema the input of the method.
+	// TODO: If a method has a single input parameter, and it's a simple type (int/string),
+	// make that the input of the method.
```
rgarcia: @kvigen left this as a TODO since I think it will involve adding a new method to the test swagger definition, and so might be a bigger change involving new tests, etc.

swagger/operation.go
```diff
@@ -7,13 +7,30 @@ import (
 	"github.com/go-openapi/spec"
 )
 
+// SingleInputSchemaParameterOp returns true if the operation has a single,
+// schema'd body input. It also returns the name of the model (without "models.").
+func SingleSchemaedBodyParameter(op *spec.Operation) (bool, string) {
+	if len(op.Parameters) == 1 && op.Parameters[0].ParamProps.Schema != nil {
+		typ, _ := TypeFromSchema(op.Parameters[0].ParamProps.Schema, false)
```
kvigen: panic if there's an error here?

swagger/swagger.go
```diff
@@ -44,8 +44,19 @@ func ImportStatements(imports []string) string {
 	if len(imports) == 0 {
 		return ""
 	}
+	remoteImports := []string{}
```
kvigen: what is this for?

rgarcia: ah sorry, on a whim I changed this so that imports follow `goimports` style of putting built-in imports at the top and third party imports at the bottom

client/genclients.go
```diff
@@ -126,15 +127,31 @@ func buildRequestCode(op *spec.Operation, method string) string {
 
 	for _, param := range op.Parameters {
 		if param.In == "body" {
-			bodyMarshalCode := fmt.Sprintf(`
+			singleInputSchema, _ := swagger.SingleSchemaedBodyParameter(op)
+			var bodyMarshalCode string
+			if singleInputSchema == true {
```
kvigen: no need for ``== true` here

swagger/operation.go
kvigen: sounds good. leave that in the doc so we can re-prioritize later?

kvigen:
lgtm other than the few minor comments!

swagger/operation.go
rgarcia: added to the doc

swagger/operation.go
rgarcia: b2acd5b

client/genclients.go
rgarcia: d736fba

PR #42, opened 2016-09-01
bugfix for single parameter methods

rgarcia:
Accidentally merged an `if` statement with an incorrect condition in the last PR. This fixes that and also adds a method that demonstrates the bug.


rgarcia:
@kvigen I might merge this now since master is in a bad state... please still review it, though!

kvigen:
lgtm. more tests!

PR #43, opened 2016-09-01
Custom string formats

kvigen:
This adds support for a `mongo-id` string format. It can be defined with a swagger.yml like:
```
authorID:
  type: string
  format: mongo-id
```

Right now it only applies to input parameters because as far as I can tell go-swagger doesn't add that type of validation to the objects defined in '#/definitions`. I think that's fine for now, as we can see how people use this and then adjust.

### DIFF
--- a/swagger.yml
+++ b/swagger.yml
@@ -33,6 +33,10 @@ paths:
           minimum: 2
           # This is a silly requirement, but let's add it to show how it would work
           multipleOf: 2
+        - name: authorID
+          in: query
+          type: string
+          format: "mongo-id"
         - name: authorization
           in: header
           type: string
@@ -168,6 +172,9 @@ definitions:
         type: integer
       name:
         type: string
+      author:
+        type: string
+        format: mongo-id
   Error:
     type: object
     properties:
--- a/swagger.yml
+++ b/swagger.yml
@@ -36,7 +36,7 @@ paths:
         - name: authorID
           in: query
           type: string
-          format: "mongo-id"
+          format: mongo-id
         - name: authorization
           in: header
           type: string
--- a/swagger/formats.go
+++ b/swagger/formats.go
@@ -1,0 +1,35 @@
+package swagger
+
+import (
+	"regexp"
+
+	"github.com/go-openapi/strfmt"
+)
+
+func InitCustomFormats() {
+	m := mongoID("")
+	strfmt.Default.Add("mongoID", &m, isMongoID)
+}
+
+var mongoRegExp = regexp.MustCompile("^[0-9a-f]{24}$")
+
+func isMongoID(str string) bool {
+	return mongoRegExp.MatchString(str)
+}
+
+type mongoID string
+
+// MarshalText turns this instance into text
+func (e mongoID) MarshalText() ([]byte, error) {
+	return []byte(string(e)), nil
+}
+
+// UnmarshalText hydrates this instance from text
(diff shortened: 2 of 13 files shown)


generated/models/inputs.go
```diff
@@ -85,6 +85,7 @@ func (i GetBooksInput) Validate() error {
 // GetBookByIDInput holds the input parameters for a getBookByID operation.
 type GetBookByIDInput struct {
 	BookID        int64
+	AuthorID      *string
```
kvigen: i thought about making this `bson.ObjectId`, but held off because it wouldn't be that in `book.go`. What do you think?

swagger.yml
```diff
@@ -33,6 +33,10 @@ paths:
           minimum: 2
           # This is a silly requirement, but let's add it to show how it would work
           multipleOf: 2
+        - name: authorID
+          in: query
+          type: string
+          format: "mongo-id"
```
rgarcia: nit: remove quotes to be consistent with `mongo-id` below

swagger/formats.go
```diff
@@ -0,0 +1,35 @@
...
+
+func InitCustomFormats() {
+	m := mongoID("")
+	strfmt.Default.Add("mongo-id", &m, isMongoID)
+}
+
+var mongoRegExp = regexp.MustCompile("^[0-9a-f]{24}$")
+
+func isMongoID(str string) bool {
+	return mongoRegExp.MatchString(str)
```
rgarcia: @agclever mentioned in another PR that the regexp condition might not be sufficient to identify valid ObjectIDs. If this is true, then we should use `bson.IsObjectIdHex`.

generated/models/inputs.go
rgarcia: Right now if you specify `date-time` or `email`, the type is a special non-string type from `strfmt`, so I could see making this an `bson.ObjectId` being consistent with that behavior. I don't feel strongly, though. We could wait and see how people use it.

swagger/formats.go
kvigen: i guess the timestamp part of it could be in the future, but that seems pathological

rgarcia: after reading the code in bson.go and `hex.Decode`, I now think the regexp check is fine

agclever: oops. seems like I am wrong there. In any case, would be nice if we used the `bson` package as the source of truth for this info. Also would be really helpful if `wag` just parsed `mongo-id` format strings into a `bson.ObjectId`.

rgarcia:
left some considerations, otherwise lgtm!

swagger/formats.go
kvigen: i thought about using a `bson.ObjectId` type and was torn. It makes things a bit easier for clients / servers, though maybe not that much easier. It also ties us more tightly to Mongo and I was a bit wary of making that even harder to change.

I definitely think this is worth re-visiting, but I'm not ready to commit yet. Let me know if you strongly disagree.

kvigen: I will just `bson.IsObjectIdHex` though. Just seems nicer.

kvigen: Oh, and one other thing. I couldn't see a way to make types marked `mongo-id` into `bson.ObjectId` in the go-swagger generated types, so our models would still be strings, and the inconsistency didn't seem great.

agclever: Makes sense. Seems like the approach with other field (like `email`) is to have a method extract the structured info. Don't have strong opinions on that.

PR #44, opened 2016-09-01
improved logging

rgarcia:
- add operation ID to all logs
- add span ID to all logs
- if returning `Default...` error types (i.e. 400 or 500), add the error text to the logs

This change will require users to upgrade to kayvee 4.1.0


kvigen:
lgtm!

PR #45, opened 2016-09-02
Write invalid code to file to debug

kvigen:
We used to write this to stdout, but I think it will be easier to debug
issues if we write it to a file because the Go error from Source has the
line number in the error. If we write the exact code to a file then the
line number will be easier to match.


rgarcia:
lgtm

PR #46, opened 2016-09-09
Ability to configure timeouts on all client requests

kvigen:
This change adds support for adding a timeout to all client requests. It is configurable in the same way that retries are configurable: both on the entire client or a particular request basis.

This seems like a generally useful thing and for now has a very simple interface.


kvigen:
cc @Christopher-Bradshaw since you were asking about this

rgarcia:
lgtm. Initially I was thinking the `context.Context` interface was good enough for this, but I see value in setting it once when creating a client.

Christopher-Bradshaw:
Very cool, thanks kv 👍

kvigen:
I also was thinking the `context.Context` interface was good enough until I got asked some questions about it.

PR #47, opened 2016-09-09
Log process metrics

kvigen:
Let's log process metrics by default so we get things like GC / heap
usage information. At some point we might consider more configurability,
including possibly https://golang.org/pkg/expvar/, but this seems like
an easy, low-impact way to start.


rgarcia:
lgtm once tests pass

PR #50, opened 2016-09-15
Enable pprof

kvigen:
I think we should enable this by default. We can go back and provide ways to turn it off later if need be.


rgarcia:
lgtm

PR #52, opened 2016-09-20
Stop cancelling requests before we're done with them

kvigen:
Before this change the call to `defer.Cancel()` (https://github.com/Clever/wag/compare/fix-timeout?expand=1#diff-4d03bde9e3d9e0386cf1997f5a89c53dL103) in the timeout doer
would cause us to cancel the underlying http request before we were
actually done processing it. For large response bodies this caused us to
get "request cancelled" errors while parsing the output.

We still want to call cancel based on the go documentation
https://github.com/golang/net/blob/master/context/go17.go#L53-L54 to
avoid temporarily leaking resources, so I decided to move the
cancellation up to the top level of code that handles http objects. This
way we only call cancel once we're actually done with the http request.


client/genclients.go
```diff
@@ -94,17 +92,27 @@ func methodCode(op *spec.Operation, basePath, method, methodPath string) string
...
 	// Add the opname for doers like tracing
 	ctx = context.WithValue(ctx, opNameCtx{}, "%s")
-	resp, err := c.requestDoer.Do(client, req.WithContext(ctx))
+	req = req.WithContext(ctx)
+	// Don't add the timeout in a "doer" because we don't want to call "defer.cancel()"
+	// until we've finished all the processing of the request object. Otherwise we'll cancel
+	// our own request before we've finished it.
+	if c.defaultTimeout != 0 {
+		ctx, cancel := context.WithTimeout(req.Context(), c.defaultTimeout)
+		defer cancel()
```
kvigen: I'm not totally sure what the Go http library does when this cancel gets called, but since we're done with the request I think it should be fairly safe. Let me know if you have any concerns. Otherwise I guess we just live with using extra memory...

test/server_test.go
```diff
@@ -269,7 +269,7 @@ func TestTimeout(t *testing.T) {
 	// Try with a global client setting
 	c = c.WithTimeout(10 * time.Millisecond)
 	err = c.HealthCheck(context.Background())
-	assert.Error(t, err)
+	require.Error(t, err)
 	assert.Equal(t, "context deadline exceeded", err.Error())
 
 }
```
rgarcia: is there an easy way to trigger the error behavior in a test? Would be nice to start locking down some of these behaviors

client/genclients.go
rgarcia: yeah seems fine to do it here

test/server_test.go
kvigen: added a test and confirmed that it failed before the change: https://github.com/Clever/wag/pull/52/commits/72dcb3eb93e3e7ceeda544c936d6e49e163c7609

kvigen:
i agree that we should re-consider the doer pattern at some point.

PR #53, opened 2016-09-21
Better logging

kvigen:
This changes a few things about our logging:
1. Instead of using the title from the swagger file, we let server
creators specify their own name. This provides a bit more flexilibity in
formatting conventions.
2. We log a counter every time the server starts to aid in debugging
server restart.

This is technically a breaking change, but given how easy the update is and how
few services are using this so far I think it's worth it.

### DIFF
--- a/gen-go/server/router.go
+++ b/gen-go/server/router.go
@@ -21,10 +21,11 @@ type Server struct {
 	// Handler should generally not be changed. It exposed to make testing easier.
 	Handler http.Handler
 	addr    string
+	l       *logger.Logger
 }
 
 // Serve starts the server. It will return if an error occurs.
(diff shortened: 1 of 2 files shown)


kvigen:
cc @mohit @natebrennand

rgarcia:
Part of me thinks this flexibility isn't necessary if we normalize the service names in the swagger.yml to lowercase and advise people that the service name should match the repo name. I think that solves the problem @natebrennand brought up in #eng-crew?

Other than that I think this is good.

kvigen:
I think you're right. Also, I started the auto-discovery client changes and realized that the service and repo name matching there would be convenient.

I will add to the docs that recommended naming convention and switch this over.

PR #54, opened 2016-09-21
Clients from discovery

kvigen:
This change adds support for creating clients from discovery environment variables. This change has two nice benefits:
1. Consumers don't have to know anything about discovery env vars
2. If we change the way discovery works we can hide it behind this interface.

### DIFF
--- a/client/genclients.go
+++ b/client/genclients.go
@@ -29,6 +29,8 @@ import (
 		"time"
 
 		"%s/models"
+
+		discovery "github.com/Clever/discovery-go"
 )
 
 var _ = json.Marshal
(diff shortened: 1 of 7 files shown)


kvigen:
cc @mohit

client/genclients.go
```diff
@@ -71,7 +83,10 @@ func (c *Client) WithTimeout(timeout time.Duration) *Client {
 	return c
 }
 
-`, packageName, s.Info.InfoProps.Title)
+`, packageName,
```
mohit: so packageName == repoName == serviceName?

mohit:
😍

client/genclients.go
kvigen: packageName contains a few extra things as well, but basically yes. I'm hoping to build this into `ark init`.

rgarcia:
lgtm!

PR #56, opened 2016-09-28
Print stack trace on panic

kvigen:
At least for now we'll still crash the process. We can continue to think
about that.

I think I missed re-generating the code for the last change, so I also have a few
of those changes mixed in.

I tested it manually. Didn't think it was worth going to the effort of writing tests for logging.


rgarcia:
Lgtm

PR #57, opened 2016-09-28
prepare for strict testing

rgarcia:
currently fails due to concurrent map access in kayvee that we
introduced and need to fix:

```
=== RUN   TestTimeout
{"ip":"127.0.0.1:50091","level":"info","method":"GET","op":"getBooks","params":"","path":"/v1/books","response-size":19001,"response-time":104822264,"source":"swagger-test","status-code":200,"title":"request-finished","via":"kayvee-middleware"}
==================
WARNING: DATA RACE
Write at 0x00c4203ad9e0 by goroutine 127:
  runtime.mapassign1()
      /usr/local/go/src/runtime/hashmap.go:442 +0x0
  github.com/Clever/wag/gen-go/server.New.func1()
      /home/vagrant/go/src/github.com/Clever/wag/gen-go/server/router.go:57 +0x17e
  net/http.HandlerFunc.ServeHTTP()
      /usr/local/go/src/net/http/server.go:1726 +0x51
  github.com/Clever/wag/vendor/github.com/gorilla/mux.(*Router).ServeHTTP()
      /home/vagrant/go/src/github.com/Clever/wag/vendor/github.com/gorilla/mux/mux.go:109 +0x187
  github.com/Clever/wag/gen-go/server.tracingMiddleware.func1()
      /home/vagrant/go/src/github.com/Clever/wag/gen-go/server/middleware.go:84 +0x4e2
  net/http.HandlerFunc.ServeHTTP()
      /usr/local/go/src/net/http/server.go:1726 +0x51
  github.com/Clever/wag/gen-go/server.panicMiddleware.func1()
      /home/vagrant/go/src/github.com/Clever/wag/gen-go/server/middleware.go:48 +0xa2
  net/http.HandlerFunc.ServeHTTP()
      /usr/local/go/src/net/http/server.go:1726 +0x51
  github.com/Clever/wag/vendor/gopkg.in/Clever/kayvee-go.v4/middleware.(*logHandler).ServeHTTP()
      /home/vagrant/go/src/github.com/Clever/wag/vendor/gopkg.in/Clever/kayvee-go.v4/middleware/middleware.go:45 +0x221
  github.com/Clever/wag/test.testContextMiddleware.func1()
      /home/vagrant/go/src/github.com/Clever/wag/test/server_test.go:218 +0x1d0
  net/http.HandlerFunc.ServeHTTP()
      /usr/local/go/src/net/http/server.go:1726 +0x51
  net/http.serverHandler.ServeHTTP()
[...]

PR #59, opened 2016-09-28
udpate to kayvee v5, turn on strict testing

rgarcia:
Fixes the data race discovered in #57 and seen in production


kvigen:
lgtm once the drone build passes

PR #60, opened 2016-09-29
Clean up interface comments

kvigen:
This removes an extra line that existed if the description was empty. It
also adds a line under the `type interface` line.


rgarcia:
lgtm

PR #61, opened 2016-09-30
Remove directories before generating code

kvigen:
To get rid of any old code / definitions


rgarcia:
Nice, lgtm. Can you also fix in swagger.mk?

kvigen:
@rgarcia which swagger.mk code?

rgarcia:
@kvigen nevermind, I was thinking of this, but it's for the js codegen https://github.com/Clever/dev-handbook/blob/1348561f36fa70913dfe3c205d7ed646925e8db6/make/swagger.mk#L49

PR #63, opened 2016-10-11
WIP constants for string enums

rgarcia:
still working on fixing the tests--the introduction of a new required param on `GetBook` causes them to fail


kvigen:
@rgarcia do we still want to prioritize this?

rgarcia:
@kvigen yeah i'd like to push this through. I'll update it and reassign when ready

PR #64, opened 2016-10-12
generate interface + mock for clients

rgarcia:
Previously we generated a `type Client struct{...}` in the `client` package. This changes the name of that struct to `WagClient` and introduces two new types:

- `type Client interface{...}`: an interface containing all client methods
- `type MockClient struct{...}`: a mock generated by mockgen


kvigen:
very nice, lgtm!

PR #65, opened 2016-10-12
Changed the default expose name to 'default' instead of 'http'

xavi-:
Related: https://github.com/Clever/template-wag/pull/6

### DIFF
--- a/client/genclients.go
+++ b/client/genclients.go
@@ -73,7 +73,7 @@ func New(basePath string) *WagClient {
 // NewFromDiscovery creates a client from the discovery environment variables. This method requires
 // the three env vars: SERVICE_%s_HTTP_(HOST/PORT/PROTO) to be set. Otherwise it returns an error.
 func NewFromDiscovery() (*WagClient, error) {
-	url, err := discovery.URL("%s", "http")
+	url, err := discovery.URL("%s", "default")
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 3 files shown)


kvigen:
is there any way we could gradually convert old services to default. Would something like

```
- name: http
  port: 80
  bind: 80
  health_check:
    type: http
    path: /health
- name: default
  port: 80
  bind: 80
  health_check:
    type: http
    path: /health
```

work?

xavi-:
Which services have been moved to wag?

kvigen:
production services currently on Wag: il-token-service, matchmaker-service, badge-generator-service, timeline, and (soon) resolve-ip.

i'm also hoping to be able to migrate existing http services over to it without having to create a new repo / launch.yml / etc..

xavi-:
There a couple of way to maintain reverse compatibility:

- Temporarily hardcode discovery-style env-vars in the upstream dependencies of the apps you listed
- Use some kind of catapult hack / workaround
- Add fallback logic to wag clients. ex:

```golang
url, err := discovery.URL("%s", "default")
if err != nil {
	url, err := discovery.URL("%s", "http")
	if err != nil {
		return err
	}
}
```

To me adding a callback to wag seems the least risky.  I could add it as part of this PR.  It should be temporary.

kvigen:
I like the fallback

xavi-:
Added fallback

kvigen:
lgtm!

you'll be happy to hear that i'm converting this code to Go templates: https://github.com/Clever/wag/pull/68/files. I'll deal with the merge conflicts

xavi-:
Sweet.  Thanks!

PR #67, opened 2016-10-14
Remove error interface

kvigen:
https://clever.atlassian.net/browse/CREW-41 for context.

This change does a couple things:
- It removes the error interface
- It changes the way we get status codes in the handler code
- It consolidates a bit of the

In the process of doing this change, I decided that we should clean up the way we access operation data. Right now we have a fair number of special cases strewn throughout the code. I didn't want to do that in this change to make it manageable, but will add a follow-up change.


rgarcia:
i like it, lgtm

PR #68, opened 2016-10-14
Encapsulate logic out of actual code generation (part 1)

kvigen:
This change starts the process of simplifying the code generation logic. We had (and still have) a lot of code that generates code based on complex analysis of the swagger operation type. Since all this logic isn't cleanly abstracted the code can be fairly hard to change. It's gotten to the point that I think it's worth refactoring.

This change has two major parts:
1. I started moving a lot more code into templates. I want to do this both because templates are fairly easy to read and also because I want to start tearing out most of the logic from the templates. Ideally the code-generation's goal is just conversion and not logic about things like which arguments should be pointers - and moving this to templates makes it more obvious when this isn't true.

2. Following the first, I want to start having most of the code never interact directly with a *swag.Operation object. If we move that behind an interface it will limit what the rest of the code can do, hopefully making it easier to maintain.

More changes are coming. I didn't have any particular reason for dividing up the change here other than this is what I could get done this afternoon.

### DIFF
--- a/client/genclients.go
+++ b/client/genclients.go
@@ -289,46 +289,50 @@ func errorMessage(err string, op *spec.Operation) string {
 `, err)
 }
 
-// TODO: Add a nice comment and ultimately move this to the (maybe an interface object with all
-// this...)
-// Response:
-//   success
-//   failure
-//   decode
-//   makePointer
-// TODO: make this into a struct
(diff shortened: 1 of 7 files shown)


rgarcia:
lgtm once drone passes

PR #69, opened 2016-10-18
Example of using additional properties

kvigen:
Thought it made sense to have at least one example of this in our sample generated code.

### DIFF
--- a/samples/gen-go/models/book.go
+++ b/samples/gen-go/models/book.go
@@ -5,8 +5,10 @@ package models
 
 import (
 	strfmt "github.com/go-openapi/strfmt"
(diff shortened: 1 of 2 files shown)


cozmo:
Hmm that's different that how I thought additionalProperties works. LGTM I think?

kvigen:
Out of curiosity, how did you expect them to work?

cozmo:
I've always thought `additionalProperties` was just a boolean, but I read up on it and seems like this is a thing you can do too.

kvigen:
Ah, yeah. That makes sense for raw json schema. For us, I think the common use case will be this - defining a map[string] type.

PR #70, opened 2016-10-19
Handle no definitions

kvigen:
This change fixes a bug I ran into when generating the appasaurus code. Before this we weren't handling the case where an operation had no success types, but also an error response.

The change itself is very straightforward. This change also adds a new pattern of having multiple swagger ymls to test off of. This makes adding new tests easier.


rgarcia:
lgtm

PR #71, opened 2016-10-19
Fix small npm install typo

ltclever:


rgarcia:
lgtm! Thanks for catching and fixing this

PR #73, opened 2016-10-20
Retry all idempotent http methods

kvigen:
Let's start doing that now so we get everyone in the habit of using the
method that maps the retry semantics they want.

Not sure why the response body close hadn't been updated earlier in the examples


rgarcia:
lgtm

PR #74, opened 2016-10-20
Move generated code to samples directory

kvigen:
Now that we're testing with multiple swagger.ymls let's move all this code to a samples directory so they don't populate the root.


rgarcia:
lgtm

PR #75, opened 2016-10-21
Don't add deprecated methods to the client

kvigen:
This makes it easier to slowly phase out old methods and makes it more
difficult for people to accidentally use them.

The actual implementation is very simple. We just skip deprecated operations
in the client.


rgarcia:
lgtm

PR #77, opened 2016-10-21
Wag patch type

kvigen:
This change adds support for a new swagger extension `x-wag-patch`. This is a convenience tag that automatically generates a new input type based on a pre-existing data type. In this case the data type has all its fields marked optional so it's easy to build 

We do this by looking through the swagger schema for any `x-wag-patch` types and then adding them to the definitions section as "magic" objects.

Right now there's an outstanding go-swagger issue that needs to be resolved before we get this completely working: https://github.com/go-swagger/go-swagger/issues/707


models/genmodels.go
```diff
@@ -40,6 +46,55 @@ func Generate(packageName, swaggerFile string, swagger spec.Swagger) error {
...
+		allDefinitions[name] = definition
+		if _, ok := definitionsToExtend[name]; ok {
+			// We calculate "Patch" + name in two places. If there becomes a third we should
+			// consolidate the logic.
+			newName := "Patch" + name
+			if _, ok := s.Definitions[newName]; ok {
+				return "", fmt.Errorf("can't apply x-wag-patch extension. Conflict with name %s", newName)
+			}
+
+			// Mark all the proeprties nullable so they show up as pointers. We do this so that
```
rgarcia: nit: "properties"

PR #78, opened 2016-10-25
Support wag-patch-type with multiple input fields

kvigen:
If you have two input fields then the generated struct doesn't have the Wag type in it.

This change updates the way we generate the input structs and the way we populate them in the server. It's a bit hacky, and exposes the fact that our code isn't quite as clean as it could be. For now, I'm living with the hack to unblock app-service work, but I'm going to go back and clean this up soon.

### DIFF
--- a/models/genmodels.go
+++ b/models/genmodels.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"os"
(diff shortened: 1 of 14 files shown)


kvigen:
ping @rgarcia

rgarcia:
lgtm

PR #79, opened 2016-10-25
Pull omitempty arrays change

kvigen:
From: https://github.com/go-swagger/go-swagger/pull/716

Also uncommented the test that was failing


rgarcia:
lgtm, but drone is failing on `glide install`...

PR #80, opened 2016-10-25
js client generation

cozmo:
This is a first pass at generating a WAG js client. 
## Supported Features
- Works with all the wag files we've seen thusfar
  - Header arguments
  - Body arguments
  - querystring arguments. 
  - Path params
- Tracing
- Timeouts (globally configurable and on a per method basis)
- Callbacks
- Promises
- Discovery based initialization
- ES5 and above supported. 
## Not currently supported
- Response type unmarshalling. Currently returns `new Error({response_keys...})` for >=400 responses and `JSON.parse(response)` for 200s. I _believe_ we should be able to add this later on in a backwards compatible way so not convinced it blocks shipping this js client. 
- Retries
- README generation
- Typescript type annotations.


goclient/genclients.go
```diff
@@ -1,4 +1,4 @@
-package client
+package goclient
```
rgarcia: nit, but what about the directories `clients/go` and `clients/js`?

jsclient/genclients.go
```diff
@@ -0,0 +1,212 @@
...
+	"github.com/Clever/wag/templates"
+	"github.com/go-openapi/spec"
+)
+
+// Generate generates a client
+func Generate(modulePath string, s spec.Swagger) error {
+	pkgName, ok := s.Info.Extensions.GetString("x-npm-package")
+	if !ok {
+		return errors.New("Must provide 'x-npm-package' in the 'info' section of the swagger.yml.") // TODO maybe validate this earlier?
+	}
```
rgarcia: 👏

jsclient/genclients.go
```diff
@@ -0,0 +1,212 @@
...
+	if err != nil {
+		return err
+	}
+
+	packageJSON, err := templates.WriteTemplate(packageJSONTmplStr, tmplInfo)
+	if err != nil {
+		return err
+	}
+
+	if err = ioutil.WriteFile(filepath.Join(modulePath, "index.ts"), []byte(indexJS), 0700); err != nil {
```
rgarcia: I'd use js file extensions, otherwise all consumers need to be running the ts transpiler

cozmo: Yea this is one of the todos in the PR description - we certainly won't be publishing TS files but for now this allows me to use/experiment/iterate in the generated code easily.

samples/gen-go-wag-patch/client/client.go
```diff
@@ -0,0 +1,169 @@
+package client
```
cozmo: I think we forgot to commit these in in a past PR - when I ran the tests they got generated.

clients/js/genjs.go
```diff
@@ -0,0 +1,249 @@
...
+
+var indexJSTmplStr = `
+const discovery = require("@clever/discovery");
+const request = require("request");
+const url = require("url");
+const opentracing = require("opentracing");
+
+// go-swagger treats handles/expects arrays in the query string to be a string of comma joined values
+// so...do that thing. It's worth noting that this has lots of issues ("what if my values have commas in them?")
+// but that's an issue with go-swagger
```
rgarcia: interesting to note that if you go to the swagger UI petstore example, there's a method `findByStatus` that takes an array in the query param, and the way it is serialized is repeating the querystring key-val: http://petstore.swagger.io/#!/pet/findPetsByStatus

samples/gen-js/index.js
```diff
@@ -0,0 +1,368 @@
...
+      this.address = options.address;
+    } else {
+      throw new Error("Cannot initialize swagger-test without discovery or address");
+    }
+    if (options.timeout) {
+      this.timeout = options.timeout
+    }
+  }
+
+  getBooks = (authors, available, state, published, snakeCase, completed, maxPages, minPages, pagesToTime, options, cb) => {
```
rgarcia: I'm really not a fan of this... how hard would it be to take in an object here, attempt to pull these values out as keys, and fail if there are extraneous keys (i.e. a super light-weight validation that might eventually be full-on json-schema validation)?

clients/js/genjs.go
```diff
@@ -0,0 +1,249 @@
...
+  }
+}
+`
+
+var methodTmplStr = `
+  {{.MethodName}} = ({{range $param := .Params}}{{$param.JSName}}, {{end}}options, cb) => {
+    if (!cb && typeof options === "function") {
+      cb = options;
+      options = undefined;
+    }
```
rgarcia: do you see us supporting callbacks forever? My main objection to supporting callbacks stems from the fact that I don't see us using callbacks forever, so eventually we will want to remove support for them here. IMO doing that (and rolling out the change everywhere) is much harder than progressively removing a handful of `bluebird.asCallback` calls in clients that *really* need callbacks.

I also don't like the fact that we are inconsistent in our use of promises / callbacks, and only supporting promises gives people an incentive to switch to promises.

clients/js/genjs.go
```diff
@@ -0,0 +1,249 @@
...
+
+    if (span) {
+      opentracing.inject(span, opentracing.FORMAT_TEXT_MAP, headers);
+      span.logEvent("{{.Method}} {{.Path}}");
+    }
+
+    const requestOptions = {
+      method: "{{.Method}}",
+      uri: this.address + "{{.BasePath}}{{.PathCode}}",
```
rgarcia: how hard would it be to use a template literal here? I don't think this lints

clients/js/genjs.go
```diff
@@ -0,0 +1,249 @@
...
+        }
+      }
+      const resolver = (data) => {
+        resolve(data);
+        if (cb) {
+          cb(null, data);
+        }
+      }
+
+      request(requestOptions, (err, response, body) => {
```
rgarcia: did you look at https://github.com/request/request-promise-native?

clients/js/genjs.go
```diff
@@ -0,0 +1,249 @@
...
+// and returns a string of javacript code such as "/path/goes/to/" + location + "/and/to/" + otherLocation
+func fillOutPath(path string) string {
+	paramRegex := regexp.MustCompile("({.+})")
+	paramNameRegex := regexp.MustCompile("{(.+)}")
+	return paramRegex.ReplaceAllStringFunc(path, func(param string) string {
+		return paramNameRegex.ReplaceAllStringFunc(param, func(paramName string) string {
+			return "\" + " + camelCase(paramName, false) + " + \""
+		})
+	})
+}
```
rgarcia: this function will keep me up at night. Can you write a test?

clients/js/genjs.go
```diff
@@ -0,0 +1,249 @@
...
+    }
+    if (options.timeout) {
+      this.timeout = options.timeout
+    }
+  }
+{{range $methodCode := .Methods}}{{$methodCode}}{{end}}
+}
+`
+
+var packageJSONTmplStr = `
```
mb-dev: `const` or `let`?

samples/gen-js-no-definitions/index.js
```diff
@@ -0,0 +1,98 @@
...
+// so...do that thing. It's worth noting that this has lots of issues ("what if my values have commas in them?")
+// but that's an issue with go-swagger
+function serializeQueryString(data) {
+  if (Array.isArray(data)) {
+    return data.join(",");
+  }
+  return data;
+}
+
+module.exports = class SwaggerTest {
```
mb-dev: should we use new ES6 exports instead of module.exports?

clients/js/genjs.go
cozmo: I feel like this is very much an issue with go-swagger, nothing about swagger itself dictates this weird serialization. 

May be worth triple checking we're correct in understanding the go-swagger behavior though..

clients/js/genjs.go
cozmo: This is go code so...neither?

clients/js/genjs.go
cozmo: Yea, but using native promises directly made it easier to support callbacks and do the unmarshalling.

clients/js/genjs.go
cozmo: I think the discussion of if Clever supports callbacks makes a lot of sense to have, but I strongly feel it is not in the scope of this repo to impart any push one way or another. _Today_ Clever uses both, and we can trivially support both in the library, _or_ we can push the burden onto all our consumers to use bluebird etc until the day when we add a lint rule that says "no callbacks ever". 

Before that day we're just making the experience of using wag clients worse for no real benefit. Take this code in the `il-api` - this is unnecessarily complex, since we are mixing callbacks and promises. https://github.com/Clever/instant-login-api/blob/master/apps/permissions.coffee#L38-L52

I'm all for a plan to fully move off of callbacks, but I'm not a fan of using these generated clients as a blunt to help us get there.

Makefile
```diff
@@ -10,21 +10,31 @@ MOCKGEN := $(GOPATH)/bin/mockgen
...
 
+cleanup:
+	rm -rf $(GOPATH)/src/$(PKG)/samples/gen-js/*
+	rm -rf $(GOPATH)/src/$(PKG)/samples/gen-js-no-definitions/*
+	rm -rf $(GOPATH)/src/$(PKG)/samples/gen-js-deprecated/*
+	rm -rf $(GOPATH)/src/$(PKG)/samples/gen-js-wag-patch/*
+	rm -rf $(PKG)/samples/gen-go/*
+	rm -rf $(PKG)/samples/gen-go-no-definitions/*
+	rm -rf $(PKG)/samples/gen-go-deprecated/*
+	rm -rf $(PKG)/samples/gen-go-wag-patch/*
```
cozmo: There's probably a better way to do this, but added this to ensure that the samples folder represents the _current_ state of the world. For an example (that I encountered) if we update the client gen to stop creating a file it may still stay committed into the samples directory.

clients/js/genjs.go
```diff
@@ -0,0 +1,249 @@
...
+	"regexp"
+
+	"github.com/Clever/wag/swagger"
+	"github.com/Clever/wag/templates"
+	"github.com/go-openapi/spec"
+)
+
+// Generate generates a client
+func Generate(modulePath string, s spec.Swagger) error {
+	pkgName, _ := s.Info.Extensions.GetString("x-npm-package")
```
cozmo: 👍  bbb7d7e

clients/js/genjs.go
```diff
@@ -0,0 +1,249 @@
...
+		return err
+	}
+
+	return nil
+}
+
+type clientCodeTemplate struct {
+	PackageName string /* Name of the npm package */
+	ClassName   string /* Name of the class to export */
+	ServiceName string /* Service's internal name (aka the github repo/discovery name) */
```
cozmo: Decided to channel my inner Abhinav and just remove the comments since I don't think they added much.

clients/js/genjs.go
cozmo: e2490eed4bdfc085c46d7a3011082820cd9d8300

samples/gen-js/index.js
cozmo: Not hard, just need to figure out exactly what we want to do.

samples/gen-js-no-definitions/index.js
cozmo: No, because we need to support this in older repos and don't want to introduce a down compilation step.

clients/js/genjs.go
cozmo: Rest easy - e8e24b8

samples/gen-js/index.js
cozmo: Here is the first pass - 64b8f71 - doesn't do any argument validation, but handles the case of 1 param vs multiple - you can see the changes in the sample code.

samples/gen-js-deprecated/package.json
```diff
@@ -0,0 +1,12 @@
+
```
rgarcia: what's up with the newlines at the top of files?

samples/gen-js-no-definitions/index.js
```diff
@@ -0,0 +1,105 @@
...
+    if (!options) {
+      options = {};
+    }
+
+    const timeout = options.timeout || this.timeout;
+    const span = options.span;
+
+    const headers = {
+
+    }
```
rgarcia: missing semicolon (maybe we should add "generated code passes lint" to the wishlist?)

samples/gen-js-no-definitions/index.js
```diff
@@ -0,0 +1,105 @@
...
+
+    if (!options) {
+      options = {};
+    }
+
+    const timeout = options.timeout || this.timeout;
+    const span = options.span;
+
+    const headers = {
+
```
rgarcia: extraneous newline

samples/gen-js-wag-patch/index.js
```diff
@@ -0,0 +1,107 @@
...
+      method: "PATCH",
+      uri: this.address + "/wagpatch",
+      json: true,
+      timeout,
+      headers,
+      body: {
+
+        "specialPatch": params.specialPatch,
+
+      },
```
rgarcia: this doesn't seem like the correct request body--compare it to how the go client does it:

https://github.com/Clever/wag/blob/e085509847bc4607c9c82aebeb3be1ca6c1aef36/samples/gen-wag-patch/client/client.go#L107
+
https://github.com/Clever/wag/blob/e085509847bc4607c9c82aebeb3be1ca6c1aef36/samples/gen-wag-patch/models/patch_data.go#L15-L28

samples/gen-js/index.js
```diff
@@ -0,0 +1,391 @@
...
+
+    const requestOptions = {
+      method: "GET",
+      uri: this.address + "/v1/books",
+      json: true,
+      timeout,
+      headers,
+      body: {
+
+      },
```
rgarcia: not sure how request handles this, but seems weird to define a body for a GET request

also feels bad in general to to define a request body for a method that doesn't take input from the request body

samples/gen-js/index.js
```diff
@@ -0,0 +1,391 @@
...
+    }
+
+    if (span) {
+      opentracing.inject(span, opentracing.FORMAT_TEXT_MAP, headers);
+      span.logEvent("GET /v1/books/{book_id}");
+    }
+
+    const requestOptions = {
+      method: "GET",
+      uri: this.address + "/v1/books/" + params.bookID + "",
```
rgarcia: use a template string?

samples/gen-js/index.js
```diff
@@ -0,0 +1,391 @@
...
+        if (response.statusCode >= 400) {
+          return rejecter(new Error(body));
+        }
+        resolver(body);
+      });
+    });
+  }
+
+  
+  getBookByID = (params, options, cb) => {
```
rgarcia: this doesn't quite follow what I laid out in slack--since this method only has one input (`bookID`), the signature should be `(bookID, options, cb)`

samples/gen-js/index.js
```diff
@@ -0,0 +1,391 @@
...
+      method: "POST",
+      uri: this.address + "/v1/books/" + params.bookID + "",
+      json: true,
+      timeout,
+      headers,
+      body: {
+
+        "newBook": params.newBook,
+
+      },
```
rgarcia: this request body also looks incorrect

Makefile
kvigen: yes, the Go generation nukes the pre-existing directory

i think everyone is going to want that to be the default, and if we force users to do it people are going to be confused. i understand the concern, but i think i would prefer encouraging people not to change the code in those packages.

samples/gen-js/index.js
cozmo: This method has a bunch more inputs than just `bookID` - https://github.com/Clever/wag/blob/e085509847bc4607c9c82aebeb3be1ca6c1aef36/samples/swagger.yml#L22-L48

samples/gen-js-no-definitions/index.js
cozmo: e7d47fb

samples/gen-js-deprecated/package.json
cozmo: e7d47fb

samples/gen-js/index.js
cozmo: Can't sadly since go doesn't let you escape backticks inside of string literals and I didn't think working around that flaw was worth the benefit.

samples/gen-js/index.js
cozmo: All these body issues were due to a misunderstanding of how body params work and our validation not catching my mistakes (fixed in #82) so I've updated the codegen to correctly handle body params in 3af5cd9

samples/gen-js-wag-patch/index.js
cozmo: All these body issues were due to a misunderstanding of how body params work and our validation not catching my mistakes (fixed in #82) so I've updated the codegen to correctly handle body params in 3af5cd9

cozmo:
@rgarcia Back to you

samples/gen-js/index.js
```diff
@@ -0,0 +1,342 @@
...
+          return rejecter(new Error(body));
+        }
+        resolver(body);
+      });
+    });
+  }
+
+  createBook = (newBook, options, cb) => {
+    const params = {};
+    params["newBook"] = params.newBook;
```
rgarcia: this code doesn't look right...

samples/gen-js/index.js
```diff
@@ -0,0 +1,342 @@
...
+          return rejecter(new Error(body));
+        }
+        resolver(body);
+      });
+    });
+  }
+
+  getBookByID2 = (id, options, cb) => {
+    const params = {};
+    params["id"] = params.id;
```
rgarcia: ditto here

rgarcia:
@cozmo couple of things but otherwise lgtm. I think we should prioritize tests that run the js client code, since most of these issues with the generated code could be caught by pretty simple tests.

clients/js/genjs.go
```diff
@@ -0,0 +1,262 @@
...
+	packageJSON, err := templates.WriteTemplate(packageJSONTmplStr, tmplInfo)
+	if err != nil {
+		return err
+	}
+
+	if err = ioutil.WriteFile(filepath.Join(modulePath, "index.js"), []byte(indexJS), 0644); err != nil {
+		return err
+	}
+
+	if err = ioutil.WriteFile(filepath.Join(modulePath, "package.json"), []byte(packageJSON), 0644); err != nil {
```
kvigen: nit: can just return `ioutil.WriteFile` here

clients/js/genjs.go
```diff
@@ -0,0 +1,262 @@
...
+module.exports = class {{.ClassName}} {
+
+  address;
+  timeout;
+
+  constructor(options) {
+    options = options || {};
+
+    if (options.discovery) {
+      this.address = discovery("{{.ServiceName}}", "default").url();
```
kvigen: in the go code we had a fallback for `http` instead of `default` for the old format. Should we add that here? https://github.com/Clever/wag/blob/master/client/genclients.go#L84

clients/js/helpers.go
```diff
@@ -0,0 +1,64 @@
+package jsclient
+
+import (
+	"bytes"
+	"regexp"
+)
+
+func camelCase(src string, capFirst bool) string {
```
kvigen: worth pulling this out into a more general helper file. I use something much primitive in another part of the code: https://github.com/Clever/wag/blob/master/swagger/parameter.go#L314

main.go
```diff
@@ -27,10 +29,14 @@ func capitalize(input string) string {
 func main() {
 
 	swaggerFile := flag.String("file", "swagger.yml", "the spec file to use")
-	packageName := flag.String("package", "", "package of the generated code")
+	goPackageName := flag.String("go-package", "", "package of the generated go code")
+	jsModulePath := flag.String("js-path", "", "path to put the js client")
```
cozmo: That'll be next.

clients/js/genjs.go
cozmo: Added.

clients/js/genjs.go
cozmo: 2c5ba9e

clients/js/helpers.go
cozmo: 2c5ba9e

PR #81, opened 2016-10-29
circuit breaker in clients

rgarcia:
@kvigen changed this up to by default be "log only" in that it will periodically log the state of the circuit. Getting this to work was a little bit tricky since hystrix-go doesn't make it easy to get the internal state of the circuit.

Here's an example kayvee log showing the state of the circuit (pulled from running `go test github.com/Clever/wag/test -test.v -run TestCircuitBreaker`):

```
{"circuitBreakerOpen":true,"errCount":149,"errPct":100,"latencyExecute":{"0":0,"25":0,"50":0,"75":0,"90":0,"95":0,"99":1,"99.5":2,"100":2},"latencyExecuteMean":0,"latencyTotal":{"0":0,"25":0,"50":0,"75":0,"90":0,"95":0,"99":1,"99.5":3,"100":3},"latencyTotalMean":0,"level":"info","reqCount":149,"rollingCountFailure":20,"rollingCountFallbackFailure":0,"rollingCountFallbackSuccess":0,"rollingCountShortCircuited":129,"rollingCountSuccess":0,"rollingCountThreadPoolRejected":0,"rollingCountTimeout":0,"source":"wag","title":"swagger-test-2a30db"}
```

The data logged is modeled after what the [`eventstream.go`](https://github.com/afex/hystrix-go/blob/39520ddd07a9d9a071d615f7476798659f5a3b89/hystrix/eventstream.go#L96-L132) optionally exposes via HTTP.

Right now this PR contains all hard-coded changes to one of the clients, after first review I'll make it auto generated.


samples/gen-go/client/circuit_logger.go
```diff
@@ -0,0 +1,239 @@
...
+				"rollingCountFailure":            uint32(m.Failures().Sum(now)),
+				"rollingCountThreadPoolRejected": uint32(m.Rejects().Sum(now)),
+				"rollingCountShortCircuited":     uint32(m.ShortCircuits().Sum(now)),
+				"rollingCountTimeout":            uint32(m.Timeouts().Sum(now)),
+				"rollingCountFallbackSuccess":    uint32(m.FallbackSuccesses().Sum(now)),
+				"rollingCountFallbackFailure":    uint32(m.FallbackFailures().Sum(now)),
+				"latencyTotal":                   generateLatencyTimings(m.TotalDuration()),
+				"latencyTotalMean":               m.TotalDuration().Mean(),
+				"latencyExecute":                 generateLatencyTimings(m.RunDuration()),
+				"latencyExecuteMean":             m.RunDuration().Mean(),
```
rgarcia: I'm realizing now that I don't log the number of concurrent requests. After looking at the code for that I think the only way to do that is via listening via HTTP to the SSE server...

samples/gen-go/client/doer.go
```diff
@@ -91,3 +99,196 @@ func (d retryDoer) Do(c *http.Client, r *http.Request) (*http.Response, error) {
...
+	d           doer
+	debug       bool
+	circuitName string
+}
+
+var circuitSSEOnce sync.Once
+
+// HystrixSSEEvent is emitted by hystrix-go via server-sent events. It describes
+// the state of a circuit.
+type HystrixSSEEvent struct {
```
kvigen: should we pull this out into a separate file?

samples/gen-go/client/doer.go
```diff
@@ -91,3 +99,196 @@ func (d retryDoer) Do(c *http.Client, r *http.Request) (*http.Response, error) {
...
+	Type                            string `json:"type"`
```
kvigen: is it worth having all these fields given that we aren't logging a bunch of them?

samples/gen-go/client/doer.go
```diff
@@ -91,3 +99,196 @@ func (d retryDoer) Do(c *http.Client, r *http.Request) (*http.Response, error) {
...
+		"rollingCountTimeout":             e.RollingCountTimeout,
+		"currentConcurrentExecutionCount": e.CurrentConcurrentExecutionCount,
+		"latencyTotalMean":                e.LatencyTotalMean,
+	})
+
+}
+
+func (d *circuitBreakerDoer) init() {
+	// Turn on the SSE port + listener once
+	circuitSSEOnce.Do(func() {
```
kvigen: this is a bit wonky. is there a chance we could make a PR against hysterix-go for an API that exposes these?

otherwise, i guess we don't have another option

test/client_test.go
```diff
@@ -124,3 +169,52 @@ func TestNewWithDiscovery(t *testing.T) {
...
+	hystrix.Flush()
+	c := client.New(testServer.URL)
+	c.SetCircuitBreakingDebug(false)
+
+	// the circuit should open after 20 consecutive failed attempts (this is
+	// the default volume threshold, after which health can open the circuit)
+	controller.down = true
+	var connAttempts int64
+	ctx := httptrace.WithClientTrace(context.Background(),
+		&httptrace.ClientTrace{
```
kvigen: nice use of client tracing :)

test/client_test.go
```diff
@@ -124,3 +169,52 @@ func TestNewWithDiscovery(t *testing.T) {
...
+		})
+	for count := 0; count < 100; count++ {
+		_, err := c.CreateBook(ctx, &models.Book{})
+		assert.Error(t, err)
+	}
+	assert.Equal(t, int64(20), connAttempts)
+
+	// we should see an attempt go through after five seconds
+	circuitOpened := time.Now()
+	for _ = range time.Tick(100 * time.Millisecond) {
```
kvigen: this can infinite loop if there's a bug. should we add some catch to prevent that

kvigen: this test takes a while. i assume we don't want to make the window configurable?

samples/gen-go/client/doer.go
```diff
@@ -91,3 +99,196 @@ func (d retryDoer) Do(c *http.Client, r *http.Request) (*http.Response, error) {
...
+	// Configure the circuit settings
+	hystrix.ConfigureCommand(d.circuitName, hystrix.CommandConfig{
+		// redundant, so need to keep this in sync with client timeout or set it high so that it's irrelevant :-/
+		Timeout: 100 * 1000,
+		// how many commands of the same type can run at the same time
+		MaxConcurrentRequests: 100,
+		// minimum number of requests needed before a circuit can be tripped due to health
+		RequestVolumeThreshold: 20,
+		// how long, in milliseconds, to wait after a circuit opens before testing for recovery
+		SleepWindow: 5000,
```
kvigen: my gut is that 5 seconds is a long time to wait to send just one request, but maybe it's better to be conservative. did we just take this number from the hysterix-go defaults?

Makefile
```diff
@@ -17,14 +17,14 @@ build: hardcoded/hardcoded.go
...
-	./bin/wag -file samples/wag-patch.yml -package $(PKG)/samples/gen-wag-patch
-	go generate ${PKG}/samples/gen-wag-patch...
+	#./bin/wag -file samples/swagger.yml -package $(PKG)/samples/gen-go
+	#go generate $(PKG)/samples/gen-go...
+	#./bin/wag -file samples/nodefinitions.yml -package $(PKG)/samples/gen-no-definitions
+	#go generate ${PKG}/samples/gen-no-definitions...
+	#./bin/wag -file samples/deprecated.yml -package $(PKG)/samples/gen-deprecated
+	#go generate ${PKG}/samples/gen-deprecated...
+	#./bin/wag -file samples/wag-patch.yml -package $(PKG)/samples/gen-wag-patch
+	#go generate ${PKG}/samples/gen-wag-patch...
```
kvigen: remember to uncomment this

kvigen:
should we do something similar for the javascript client?

samples/gen-go/client/doer.go
rgarcia: @kvigen I filed an issue https://github.com/afex/hystrix-go/issues/56

samples/gen-go/client/doer.go
rgarcia: I was worried about the case where we may want to log a different set of metrics, and how it might be useful to know what is available... For now I'll remove the ones were not using

test/client_test.go
rgarcia: good idea... i'll expose a way to change the circuit settings to something other than the defaults and then use it in this test

rgarcia:
@kvigen can you take one more look? I've responded to your comments and also made the code autogenerated.

rgarcia:
also yes, we should do this for the js client, but I think I'm going to tackle retries in both js/go first

kvigen:
doing retries first in javascript makes sense to me

test/client_test.go
```diff
@@ -124,3 +170,69 @@ func TestNewWithDiscovery(t *testing.T) {
...
+	// bring the server back up, and we should see successes after another
+	// two seconds, for a total of 4 seconds.
+	controller.down = false
+	for _ = range time.Tick(100 * time.Millisecond) {
+		_, err := c.CreateBook(ctx, &models.Book{})
+		if err == nil {
+			assert.WithinDuration(t, time.Now(), circuitOpened,
+				4*time.Second+500*time.Millisecond)
+			break
+		}
```
kvigen: should we have the same check as above to make sure it doesn't infinite loop?

test/client_test.go
```diff
@@ -124,3 +170,69 @@ func TestNewWithDiscovery(t *testing.T) {
...
+	c.SetCircuitBreakerDebug(false)
+	c.SetCircuitBreakerSettings(client.CircuitBreakerSettings{
+		MaxConcurrentRequests:  client.DefaultCircuitBreakerSettings.MaxConcurrentRequests,
+		RequestVolumeThreshold: 1,
+		SleepWindow:            2000,
+		ErrorPercentThreshold:  client.DefaultCircuitBreakerSettings.ErrorPercentThreshold,
+	})
+
+	for k, v := range hystrix.GetCircuitSettings() {
+		log.Print(k, *v)
```
kvigen: remove this?

samples/gen-go/client/doer.go
kvigen: true, that seems reasonable to me

_hardcoded/doer.go
```diff
@@ -91,3 +100,146 @@ func (d retryDoer) Do(c *http.Client, r *http.Request) (*http.Response, error) {
...
+		"rollingCountThreadPoolRejected":  e.RollingCountThreadPoolRejected,
+		"rollingCountTimeout":             e.RollingCountTimeout,
+		"currentConcurrentExecutionCount": e.CurrentConcurrentExecutionCount,
+		"latencyTotalMean":                e.LatencyTotalMean,
+	})
+
+}
+
+func (d *circuitBreakerDoer) init() {
+	// Turn on the SSE port + listener once
```
rgarcia: 877a81d

test/client_test.go
rgarcia: whoops 546d9c0

test/client_test.go
rgarcia: a9fcabf

PR #82, opened 2016-10-30
Validation fixes

cozmo:
Previously we weren't validating the sample ymls very well. This meant we had invalid yml files which caused me some confusion with the js codegen. 

I ran `go-swagger validate` against each sample file and fixed any errors I found. Then I added a call to go-swagger's validation code in our validate function so now validation is 100% contained within `wag` instead of having to put a goswagger validate call in the Makefile(s). 

I think I've seen a few tickets about consolidating the validation logic so this should accomplish those as well.

### DIFF
--- a/main.go
+++ b/main.go
@@ -40,7 +40,7 @@ func main() {
 	}
 	swaggerSpec := *doc.Spec()
 
-	if err := validation.Validate(swaggerSpec); err != nil {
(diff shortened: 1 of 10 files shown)


cozmo:
Assigning to @kvigen since @rgarcia got my js codegen PR :p

rgarcia:
lgtm

PR #83, opened 2016-10-31
publish to github not docker

cozmo:


Makefile
```diff
@@ -11,8 +14,7 @@ $(MOCKGEN):
 	go get -u github.com/golang/mock/mockgen
 
 build: hardcoded/hardcoded.go
-	# disable CGO and link completely statically (this is to enable us to run in containers that don't use glibc)
-	CGO_ENABLED=0 go build -installsuffix cgo -o bin/wag
+	go build -installsuffix cgo -o bin/wag
```
rgarcia: `-installsuffix cgo` shouldn't be required anymore

rgarcia:
lgtm

kvigen:
i assume we did this because we don't need java anymore so the docker container wasn't adding much value?

cozmo:
Yep

PR #85, opened 2016-11-01
Use go-errors to get error backtraces.

samfishman:
This modifies the generated handler code to include a stacktrace in the kayvee error logging when the error is an instance of https://github.com/go-errors/errors.


kvigen:
lgtm, thanks for adding this

kvigen:
we should add a note to the readme about this

kvigen:
👍  on the readme

PR #87, opened 2016-11-01
Add wag --version

cozmo:
We were [already](https://github.com/Clever/wag/blob/master/Makefile#L50) writing the version in during build - now expose it with `wag -version`.


main.go
```diff
@@ -26,12 +26,19 @@ func capitalize(input string) string {
...
+
 func main() {
 
 	swaggerFile := flag.String("file", "swagger.yml", "the spec file to use")
 	goPackageName := flag.String("go-package", "", "package of the generated go code")
 	jsModulePath := flag.String("js-path", "", "path to put the js client")
+	versionFlag := flag.Bool("version", false, "print the wag version and exit")
 	flag.Parse()
+	if *versionFlag {
+		fmt.Println(version)
```
kvigen: where do we set the `version` variable?

kvigen: oh, i see... that's weird...

kvigen:
lgtm

PR #88, opened 2016-11-03
Remove wag patch

kvigen:
As discussed in the JIRA https://clever.atlassian.net/browse/CREW-93 it turned out that w-wag-patch wasn't as useful as originally anticipated.

- The fields in the update didn't exactly match the fields in the create
- Wag-patch, at least as currently implemented, didn't support nested types well
- It's somewhat magical and can be confusing to users

Given that we now have a better workaround than we though we did when we first implemented it (the x-isnullable workaround) I think we should remove the magic for now and make the swagger.yml more explicit. We can add this back later if the demand materializes.


rgarcia:
lgtm

PR #89, opened 2016-11-04
WIP exponential backoff for node/go clients

rgarcia:
also first node client test

currently this branch just contains direct modifications to the generated client code. I will autogenerate the code after I get the 👍 from @cozmo and @kvigen.


samples/gen-js/index.js
```diff
@@ -13,6 +13,28 @@ function serializeQueryString(data) {
   return data;
 }
 
+const defaultRetryPolicy = {
```
cozmo: I think we want to export this right?

cozmo: `module.exports.RetryPolicies = {defaultRetryPolicy};`?

samples/gen-js/index.js
```diff
@@ -88,15 +113,25 @@ module.exports = class SwaggerTest {
...
-      request(requestOptions, (err, response, body) => {
-        if (err) {
-          return rejecter(err);
-        }
-        if (response.statusCode >= 400) {
-          return rejecter(new Error(body));
-        }
-        resolver(body);
-      });
+      const retryPolicy = this.retryPolicy || options.retryPolicy || defaultRetryPolicy;
```
cozmo: Can I pass `null` as a retryPolicy? I feel like in the current model I can't.

cozmo: But that is something I would assume I'd be able to do.

samples/gen-js/test/test.js
```diff
@@ -0,0 +1,35 @@
...
+
+const Client = require("../index");
+
+describe("retries", function() {
+  it("exponential backoff by default", function(done) {
+    const client = new Client({address: "http://localhost:8000"});
+    let requestCount = 0;
+    let requestTimes = [];
+    const scope = nock("http://localhost:8000")
+            .get("/v1/books")
```
cozmo: indents look a little funky here.

samples/gen-js/test/test.js
```diff
@@ -0,0 +1,35 @@
+const assert = require("assert");
```
cozmo: These tests look real nice - can we make them run as part of `make test` in the wag repo?

samples/gen-js/index.js
rgarcia: 👍  58eee55

samples/gen-js/index.js
rgarcia: I'm assuming by `null` you're trying to communicate "no retries"? If so, I think I'd rather export a `NoRetryPolicy` that you can pass (this is what I do with the Go client)

samples/gen-js/test/test.js
rgarcia: yeah, will set that up once this is all auto-generated. Right now it'll all get overwritten in CI

samples/gen-js/index.js
rgarcia: check this out 3e5d27b

samples/gen-js/index.js
```diff
@@ -341,3 +385,8 @@ module.exports = class SwaggerTest {
     });
   }
 }
+
+module.exports.RetryPolicies = {
+  Default: defaultRetryPolicy,
+  No: noRetryPolicy,
```
cozmo: s/no/none/?

samples/gen-go/client/doer.go
```diff
@@ -58,45 +59,83 @@ func (d tracingDoer) Do(c *http.Client, r *http.Request) (*http.Response, error)
...
 	}
+	return ret
+}
 
-	// If the request can't be retried then just return immediately. We retry all idempotent
-	// http requests. The only two that can't be retried are post and patch
-	if r.Method == "POST" || r.Method == "PATCH" {
-		return resp, err
+// Retry will retry non-POST, non-PATCH requests that 5XX.
+// TODO: It does not currently retry any errors returned by net/http.Client's `Do`.
```
kvigen: fyi, here's the Go transport retrying I mentioned https://github.com/golang/go/blob/master/src/net/http/transport.go#L403-L450

test/client_test.go
```diff
@@ -93,7 +95,10 @@ func TestDefaultClientRetries(t *testing.T) {
 	c := client.New(testServer.URL)
 	_, err := c.GetBooks(context.Background(), &models.GetBooksInput{})
 	assert.NoError(t, err)
-	assert.Equal(t, 2, controller.getCount)
+	backoff := controller.getTimes[1].Sub(controller.getTimes[0])
+	if backoff < 90*time.Millisecond || backoff > 110*time.Millisecond {
```
kvigen: i recently learned about https://github.com/stretchr/testify/blob/master/assert/assertions.go#L722-L727 which makes this assert a little nicer

test/client_test.go
```diff
@@ -93,7 +95,10 @@ func TestDefaultClientRetries(t *testing.T) {
 	c := client.New(testServer.URL)
 	_, err := c.GetBooks(context.Background(), &models.GetBooksInput{})
 	assert.NoError(t, err)
-	assert.Equal(t, 2, controller.getCount)
+	backoff := controller.getTimes[1].Sub(controller.getTimes[0])
```
kvigen: should we add a test for more than one retry?

samples/gen-js/test/test.js
kvigen: we're still thinking of writing some tests where we don't mock the server along with these, right?

samples/gen-go/client/doer.go
rgarcia: interesting that they only retry GET, HEAD, OPTIONS, TRACE https://github.com/golang/go/blob/117c9c35cde8444d26db91bfa8346dde252989b1/src/net/http/request.go#L1252-L1260

rgarcia: the rest of the retry logic there is pretty cryptic to me... do you understand what it's doing?

samples/gen-js/test/test.js
rgarcia: yeah, I think we should eventually write some full E2E tests, but I'm not sure the best way to structure them...

samples/gen-js/index.js
rgarcia: bc95207

test/client_test.go
rgarcia: 43740b6

test/client_test.go
rgarcia: 43740b6

rgarcia:
@cozmo finished auto generating everything and moved the js tests to their own directory--want to give it a final review?

.npmrc_drone
```diff
@@ -0,0 +1,5 @@
+ca = null
+//registry.npmjs.org/:_authToken=${npm_auth_token}
```
cozmo: Only minor comment is I'd love to `s/npm_auth_token/npm_token/` to be more consistent with the npm changes I made to the 60 repos over the last week.

cozmo:
@rgarcia Left one minor comment but otherwise LGTM.

samples/gen-go/client/doer.go
kvigen: only kind of. i'm still digging through the Go http library. will let you know if i understand it better as i read more

PR #90, opened 2016-11-04
Fix array encoding

cozmo:
Fixes go server gen to accept arrays in query strings in the format

```
/path?key=val1&key=val2&key=val3
```

Instead
```
/path?key=val1,val2,val3
```

Also update the JS clients to not hack themselves to support this backwards model. 

## TODO
- [x] Verify this actually works by doing E2E testing of some form.

### DIFF
--- a/VERSION
+++ b/VERSION
@@ -1,2 +1,2 @@
-0.2.2
-- Fix release process to include v prefix
+0.3.0
+- Don't comma encode arrays in querystring params - instead use multiple querystring items.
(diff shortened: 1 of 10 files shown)


cozmo:
Suggest viewing with whitespace insensitive diff so you can see _Kyle's basket of sprintfs_ in it's full glory.

kvigen:
lgtm

cozmo:
Tested this locally hitting a generated server with the clients. Seems to work well. Certainly backwards incompatible so bumping to 0.3.0.

PR #91, opened 2016-11-04
Better error types

kvigen:
This change adds support for more error features:
- You can define global error types in your yml
- You can define your own 400 + 500 responses as long as they have a Msg field
- If your operations don't have a 400 or 500 defined then they implicitly get the global ones

There are several parts of the change:
- Modify the input swagger yml to validate the error types and add the implicit ones
- No longer hard-code DefaultInternalError and DefaultBadRequest since they can be overridden
- More info about errors in the readme
- Removed the nodefinitions test code since it's now impossible to have a yml without at least an error definition.
- All the generated code changes


samples/gen-go-deprecated/models/outputs.go
```diff
@@ -1,24 +1,30 @@
 package models
 
-// DefaultInternalError represents a generic 500 response.
-type DefaultInternalError struct {
-	Msg string `json:"msg"`
-}
+// BadRequest defines a response type.
+// Bad Request
+type BadRequest Error
```
kvigen: changed the name from `DefaultBadRequest` to `BadRequest`. This isn't backwards compatible, but I think I like it better, and since we're already requiring users to change their yml, this seems reasonable.

samples/gen-go/server/handlers.go
```diff
@@ -569,10 +604,18 @@ func statusCodeForHealthCheck(obj interface{}) int {
...
 
-	case models.DefaultBadRequest:
+	case *models.BadRequest:
 		return 400
-	case models.DefaultInternalError:
+
+	case *models.InternalError:
 		return 500
+
+	case models.BadRequest:
```
kvigen: note: added support for both the pointer and non-pointer form just to protect against the user doing something unexpected - since that wouldn't be caught at compile time

README.md
```diff
@@ -17,6 +17,29 @@ generate: wag-generate-deps
...
+
+definitions:
+  Error:
+    type: object
+    properties:
+      msg:
+        type: string
+```
+
+If your yml doesn't define a 400 or a 500 response for any of your operations, Wag will implicitly set it to the globally defined response.
```
rgarcia: how does wag identify these? If it's by name, e.g. `BadRequest`, `InternalError`, then I'd mention that here.

README.md
```diff
@@ -52,19 +75,29 @@ logger.FromContext(ctx).Info(...)
 
   You should use this logger for all logging within your controller implementation.
 
-* **Errors**. Errors returned from your controller are logged by the
+* **Errors**.
+  * Wag supports three types of errors
+    * Global error response types: "models.{{.ResponseName}}"
```
rgarcia: it's not immediately clear what this means--what is `ResponseName` here?

swagger/transforms.go
```diff
@@ -0,0 +1,159 @@
...
+	"fmt"
+
+	"github.com/go-openapi/jsonreference"
+	"github.com/go-openapi/spec"
+)
+
+// TransformErrors transforms the errors of the swagger spec object. This means
+// both verifying that the required errors exist, and also adding the 400 / 500
+// responses to any operation that doesn't have them defined.
+func TransformErrors(s spec.Swagger) error {
```
rgarcia: What is being `Transform`ed? The naming of this function feels off to me

README.md
kvigen: yup, will add

swagger/transforms.go
kvigen: the `spec.Swagger` is get transformed because we're adding the default 400 + 500 responses. That said, I agree that it doesn't describe the function very well. maybe i should name it `ValidateErrors` since that's much more of what it does then add the extra responses.

rgarcia:
changes look good!

PR #92, opened 2016-11-04
remove wrapper struct for methods with a single string parameter

rgarcia:
For a method spec like this:

```yaml
  /books2/{id}:
    get:
      operationId: getBookByID2
      description: Retrieve a book
      parameters:
        - name: id
          in: path
          required: true
          type: string
          pattern: "^[0-9a-f]{24}$"
      responses:
        200:
          description: OK response
          schema:
            $ref: '#/definitions/Book'
```

We used to generate a client / server interface + input struct like this

```go
type ServerAndClientInterface interface {
    GetBookByID2(ctx context.Context, i *models.GetBookByID2Input) (*models.Book, error)
}

type GetBookByID2Input struct {
	ID string
}
```

This PR removes the input struct and uses the basic type (string) instead:

```go
type ServerAndClientInterface interface {
    GetBookByID2(ctx context.Context, id string) (*models.Book, error)
}
```

https://clever.atlassian.net/browse/CREW-12

### DIFF
--- a/models/genmodels.go
+++ b/models/genmodels.go
@@ -4,11 +4,13 @@ import (
 	"bytes"
 	"fmt"
 	"os"
+	"strings"
 
 	"github.com/go-openapi/spec"
 
 	"github.com/Clever/wag/swagger"
 	"github.com/Clever/wag/templates"
+	"github.com/Clever/wag/utils"
 
 	"github.com/go-swagger/go-swagger/generator"
 )
@@ -74,7 +76,7 @@ var _ = strfmt.NewFormats
 			// operation that has a single, schema'd input.
 			// The input to these will be the model generated for
 			// the schema.
-			if ssbp, _ := swagger.SingleSchemaedBodyParameter(op); ssbp {
+			if singleSchemaedBodyParameter, _ := swagger.SingleSchemaedBodyParameter(op); singleSchemaedBodyParameter {
 				continue
 			}
 			if err := printInputStruct(&g, op); err != nil {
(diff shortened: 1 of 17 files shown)


Makefile
```diff
@@ -21,10 +21,10 @@ test: build generate $(PKGS)
...
 	./bin/wag -file samples/swagger.yml -go-package $(PKG)/samples/gen-go -js-path $(GOPATH)/src/$(PKG)/samples/gen-js
 	go generate $(PKG)/samples/gen-go...
-	./bin/wag -file samples/nodefinitions.yml -go-package $(PKG)/samples/gen-go-no-definitions -js-path $(GOPATH)/src/$(PKG)/samples/gen-js-no-definitions
-	go generate ${PKG}/samples/gen-go-no-definitions...
-	./bin/wag -file samples/deprecated.yml -go-package $(PKG)/samples/gen-go-deprecated -js-path $(GOPATH)/src/$(PKG)/samples/gen-js-deprecated
-	go generate ${PKG}/samples/gen-go-deprecated...
+	# ./bin/wag -file samples/nodefinitions.yml -go-package $(PKG)/samples/gen-go-no-definitions -js-path $(GOPATH)/src/$(PKG)/samples/gen-js-no-definitions
+	# go generate ${PKG}/samples/gen-go-no-definitions...
+	# ./bin/wag -file samples/deprecated.yml -go-package $(PKG)/samples/gen-go-deprecated -js-path $(GOPATH)/src/$(PKG)/samples/gen-js-deprecated
+	# go generate ${PKG}/samples/gen-go-deprecated...
```
kvigen: reminder to add these back

models/genmodels.go
```diff
@@ -118,11 +120,20 @@ func printInputStruct(g *swagger.Generator, op *spec.Operation) error {
...
+		g.Printf("// Validate%sInput returns an error if the input parameter doesn't\n",
+			capOpID)
+		g.Printf("// satisfy the requirements in the swagger yml file.\n")
+		g.Printf("func Validate%sInput(%s string) error{\n", capOpID, paramName)
+	} else {
+		capOpID := swagger.Capitalize(op.ID)
+		g.Printf("// Validate returns an error if any of the %sInput parameters don't satisfy the\n",
+			capOpID)
+		g.Printf("// requirements from the swagger yml file.\n")
+		g.Printf("func (i %sInput) Validate() error{\n", capOpID)
```
kvigen: should we make the naming consistent with the new string format?

server/genserver.go
```diff
@@ -284,13 +284,16 @@ func generateOperationHandler(op *spec.Operation) (string, error) {
 	}
 
 	singleInputOp, _ := swagger.SingleSchemaedBodyParameter(op)
```
kvigen: we should probably rename this variable

models/genmodels.go
rgarcia: f28aaefc6d03465cdf951edd6f55288cd3f9d31f

server/genserver.go
rgarcia: f28aaefc6d03465cdf951edd6f55288cd3f9d31f

models/genmodels.go
rgarcia: err, sorry, not sure what you mean by this?

kvigen: sorry, wasn't very clear. I meant whether we should we write the function as
```
func Validate%sInput(i %sInput) error
```

so it's like the single string path case

rgarcia: I think I prefer to keep it as-is (method on the type). As written I think the standalone / package-level `Validate*` method introduced here is not ideal... it's one of the downsides of not having the struct anymore.

PR #93, opened 2016-11-04
Support error types with fields

kvigen:
https://clever.atlassian.net/browse/CREW-55

This change has three parts:
- Sort the global responses (so every build doesn't have unimportant diffs)
- Actually parse the result as expected
- Refactor this code. This included changing the returned type of errors from the object to a pointer to the object. I think this is probably a good change this it matches the success type returned, but I'm open to other thoughts. Slices are still returned as non-pointers.

### DIFF
--- a/samples/gen-go-errors/client/client.go
+++ b/samples/gen-go-errors/client/client.go
@@ -41,7 +41,11 @@ var _ Client = (*WagClient)(nil)
 func New(basePath string) *WagClient {
 	base := baseDoer{}
 	tracing := tracingDoer{d: base}
+<<<<<<< HEAD
 	retry := retryDoer{d: tracing, retryPolicy: DefaultRetryPolicy{}}
+=======
+	retry := retryDoer{d: tracing, defaultRetries: 1}
+>>>>>>> Generated
 	circuit := &circuitBreakerDoer{
 		d:     &retry,
 		debug: true,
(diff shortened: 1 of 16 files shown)


samples/gen-go-errors/client/client.go
```diff
@@ -179,25 +180,48 @@ func (c *WagClient) GetBook(ctx context.Context, i *models.GetBookInput) error {
...
+			return &models.InternalError{Msg: err.Error()}
+		}
+		return &output
+
 	case 404:
+
 		var output models.NotFound
-		return output
+		// Any errors other than EOF should result in an error. EOF is acceptable for empty
+		// types.
```
kvigen: i'm okay with this for now, but i would love to clean up

samples/gen-go-errors/client/client.go
```diff
@@ -68,10 +69,10 @@ func NewFromDiscovery() (*WagClient, error) {
...
 }
 
-// WithRetries returns a new client that retries all GET operations until they either succeed or fail the
-// number of times specified.
-func (c *WagClient) WithRetries(retries int) *WagClient {
-	c.retryDoer.defaultRetries = retries
+// WithRetryPolicy returns a new client that will use the given retry policy for
+// all requests.
+func (c *WagClient) WithRetryPolicy(retryPolicy RetryPolicy) *WagClient {
+	c.retryDoer.retryPolicy = retryPolicy
```
rgarcia: does this branch need a rebase?

kvigen: yup, hadn't run the latest `make generate` on the old branch

PR #94, opened 2016-11-08
Always return errors defined in definitions

kvigen:
We think having a more consistent way of returning errors will make it easier for users to understand what's going on. It also removes some cruft from the code and some superfluous data types.

This change does this by doing a couple things:
- Following error type references all the way to the schema
- Generating `Error() string` methods for all error types
- Enforcing that all error types have a `Msg` field
- Cleaning up the related code


models/genmodels.go
```diff
@@ -324,16 +266,50 @@ var typeTemplate = `
...
 	}
-	{{end}}
 `
+
+// generateErrorMethods finds all responses all error responses and generates an error
+// method for them.
+func generateErrorMethods(s *spec.Swagger) (string, error) {
+	errorTypes := stringset.New()
+
+	for _, pathKey := range swagger.SortedPathItemKeys(s.Paths.Paths) {
```
kvigen: you can pull out this code into a helper function if it's useful for the JS client

kvigen:
cc @rgarcia

README.md
```diff
@@ -77,11 +84,11 @@ logger.FromContext(ctx).Info(...)
...
   * Wag supports three types of errors
-    * Global error response types: "models.{{.ResponseID}}"
-    * Response types for a specific operation: "models.{{.OperationID}}{{.StatusCode}}Output"
-    * Undefined error types: anything that implements the Error interface
+    * Global error response types
+    * Response types for a specific operation
+    * Undefined error types
 
-  * Any of the three can be returned from a controller. If an undefined error type is returned then Wag will convert the error to a 500. Otherwise Wag will use the status code defined in the yml responses section.
+  * Any of the three can be returned from a controller. To return a global or response specific error type use return a pointer to the model defintion for that error type. To return an undefined error type, return anything that implements the Error interface. Wag automatically converts undefined errors into the 500 error type.
```
cozmo: Worth adding a note about non error return types here for completeness?

models/genmodels.go
cozmo: Nice 👌

cozmo:
Although looks like we need to `glide install` stringset.

README.md
kvigen: https://github.com/Clever/wag/pull/94/commits/44e8c69234de754323c75058a37d9f1534274891

PR #98, opened 2016-11-08
JS Error Types

cozmo:
Due to the simplification we discussed earlier js types were...very easy. 

This branch adds `module.exports.Errors` which export all the possible error classes. When we get a response error back we turn it into one of these error types. This also improves the general response handling to support functions that don't return any data for example. 

Also fixes a bunch of random lint stuff I caught while linting the types.

### DIFF
--- a/clients/js/genjs.go
+++ b/clients/js/genjs.go
@@ -5,7 +5,9 @@ import (
 	"io/ioutil"
 	"path/filepath"
 	"regexp"
+	"strings"
 
+	"github.com/Clever/go-utils/stringset"
 	"github.com/Clever/wag/swagger"
 	"github.com/Clever/wag/templates"
 	"github.com/Clever/wag/utils"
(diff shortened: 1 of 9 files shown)


clients/js/genjs.go
```diff
@@ -241,6 +264,14 @@ type paramMapping struct {
 	Required bool
 }
 
+type responseMapping struct {
+	StatusCode int
+	Name       string
+	IsError    bool
+	IsRichType bool
```
kvigen: is this used?

clients/js/genjs.go
```diff
@@ -213,17 +226,27 @@ var methodTmplStr = `
...
+          switch (response.statusCode) {
+            {{ range $response := .Responses }}case {{ $response.StatusCode }}:{{if $response.IsError }}
+              rejecter(new Errors.{{ $response.Name }}(body || {}));
+            {{else}}{{if $response.IsNoData}}
+              resolver();
+            {{else}}
+              resolver(body);
+            {{end}}{{end}}  break;
+            {{end}}default:
+              rejecter(new Error("Recieved unexpected statusCode " + response.statusCode));
```
kvigen: in go we turn this into an internal error `new InternalError(...`, not sure if we want to do that in JS too

clients/js/genjs.go
```diff
@@ -305,3 +354,50 @@ func methodCode(op *spec.Operation, method, basePath, path string) (string, erro
...
+    super(body.message);
+    for (const k of Object.keys(body)) {
+      this[k] = body[k];
+    }
+  }
+};
+{{ end }}
+`
+
+func generateTypesFile(s spec.Swagger) (string, error) {
```
kvigen: do we want to factor this out? i can also do this in a follow up

PR #99, opened 2016-11-08
Msg to message

kvigen:
We decided to do this for two reasons:
1. Javascript errors have to have a message field so it seems reasonable to match it here
2. It seems nicer to not use an abbreviation


cozmo:
LGTM.

cozmo:
Remember to bump the version during deploy.

samfishman:
There's still a reference to `msg` in the README (L20... github won't let me leave a comment on that line).

PR #100, opened 2016-11-08
Fix error message typo

kvigen:
Thanks for catching this


samfishman:
LGTM (also see my comment on your last PR for another typo 😛 )

kvigen:
man, i swear i grepped for all references. Cool, will update that here too

PR #101, opened 2016-11-08
Only support one success type

kvigen:
The way we handle multiple success type is fairly wonky right now: we create a wrapper struct that has a interface that all the success types need to implement. Though this isn't the end of the world, it isn't great either, and since we don't have any use cases for it right now, we decided to remove it. If we ever want to support in the future then we can. And hopefully then we'll have a better idea of how to support it.

### DIFF
--- a/swagger/transforms.go
+++ b/swagger/transforms.go
@@ -8,10 +8,12 @@ import (
 	"github.com/go-openapi/spec"
 )
 
-// ValidateErrors transforms the errors of the swagger spec object. This means
-// both verifying that the required errors exist, and also adding the 400 / 500
-// responses to any operation that doesn't have them defined.
-func ValidateErrors(s spec.Swagger) error {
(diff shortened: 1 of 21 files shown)


swagger/transforms.go
```diff
@@ -61,6 +66,10 @@ func ValidateErrors(s spec.Swagger) error {
 				}
 			}
 
+			if successResponseCount > 1 {
+				return fmt.Errorf("can only define one success type (statusCode < 400) for %s", op.ID)
+			}
+
```
cozmo: Want to just add validation that we don't have any 3XX responses? We can always remove later if we see a logical reason to have one.

cozmo:
This was a good simplication. The generated code LGTM and the generation code didn't change much besides getting clearer so LGTM besides one quick comment/question.

swagger/transforms.go
kvigen: sure, that's easy enough

PR #103, opened 2016-11-08
Remove io.EOF checking

kvigen:
This was wonky error checking that we added because we didn't always
have types and so we sometimes serialized the data to the empty string.
Now that this case isn't possible (as of https://github.com/Clever/wag/pull/101) let's remove the error checking since
EOF should be an error.


cozmo:
LGTM

PR #104, opened 2016-11-09
Standardize client APIs

kvigen:
I was doing a last pass on the APIs to make sure they were ready for v1 and I noticed that sometimes we use `With*` and sometimes `Set*` for setting client fields. I think I like Set better, so I standardized on that. Let me know if you would rather go with `With*`

I also removed JoinByFormat now that we don't encode array values that way any more.


rgarcia:
lgtm

PR #105, opened 2016-11-10
Don't wrap client errors in a 500

kvigen:
This allows users to do things like check for the context timeout error.
It's also was a little misleading to describe those as 500s given that
the server didn't actually return a 500.


kvigen:
cc @mohit this makes Wag return the right error type for timeouts and other errors that happen client side.

rgarcia:
lgtm

PR #106, opened 2016-11-10
Temporarily change the default retry policy

kvigen:
Until we turn on circuit breakers by default lets reduce the number of
retries so that we don't overload our system with retries.


rgarcia:
@kvigen can you change the name of `DefaultRetryPolicy` to `ExponentialRetryPolicy` now that it's not the default? Also, once the JSDoc stuff is merged can you make the corresponding change to the JS client?

otherwise lgtm!

kvigen:
@rgarcia addressed the comments. Do you mind taking one more pass?

Makefile
```diff
@@ -19,7 +19,7 @@ build: hardcoded/hardcoded.go
 test: build generate $(PKGS) js-tests
 
 js-tests:
-	cd test/js && npm install && npm test
+	cd test/js && rm -rf node_modules && npm install && npm test
```
rgarcia: this might slow things down when developing locally--is it absolutely necessary?

clients/js/genjs.go
```diff
@@ -120,6 +120,24 @@ const defaultRetryPolicy = {
 };
 
 /**
+ * Use this retry policy to retry a request once.
+ * @alias module:{{.ServiceName}}.RetryPolicies.Single
+ */
+const singleRetryPolicy = {
+  backoffs() {
+    return [1000]
```
rgarcia: missing semicolon

clients/js/genjs.go
```diff
@@ -120,6 +120,24 @@ const defaultRetryPolicy = {
...
+  },
+  retry(requestOptions, err, res) {
+    if (err || requestOptions.method === "POST" ||
+        requestOptions.method === "PATCH" ||
+        res.statusCode < 500) {
+      return false;
+    }
+    return true;
+  },
+}
```
rgarcia: missing semicolon

Makefile
rgarcia: ahh i see... I think it's fine to leave as is, and if it slows things down significantly we can try just removing node_modules/swagger-test

PR #107, opened 2016-11-10
Generate docs for the JS client

rgarcia:
This adds JSDoc comments to the generated js code, and runs `jsdoc2md` to create a README. You can see an example here: https://github.com/Clever/wag/tree/CREW-95-js-docs/samples/gen-js

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -21,12 +21,19 @@ test: build generate $(PKGS) js-tests
 js-tests:
 	cd test/js && npm install && npm test
 
-generate: hardcoded/hardcoded.go $(MOCKGEN)
+jsdoc2md:
+	hash npm 2>/dev/null || (echo "Could not run npm, please install node" && false)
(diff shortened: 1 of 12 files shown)


cozmo:
@rgarcia I think this looks pretty good. I made some tweaks in 917b3f4 - reposted them below. 

- Removed documentation for individual methods of retryPolicies - I thought they were confusing for the user since they're not calling/using them
- Fixed the descriptions of `address` and `discovery` which were incorrect
- Added `options` to the method docs.
- Added links for `retryPolicy` arguments to take the user to the documentation of which retryPolicies they can use
- Added links for the `Opentracing.Span` arguments to take the user to the opentracing docs
- Pull param descriptins through from swagger file

With these changes I'd say this LGTM, but I'd say you should check out the new README and see what you think - https://github.com/Clever/wag/tree/CREW-95-js-docs/samples/gen-js

rgarcia:
Nice. Commit lgtm, so I will merge

PR #108, opened 2016-11-10
Sort schema properties

kvigen:
Without this the generated docs randomly changed each time I built


rgarcia:
nice, I noticed this too. lgtm

PR #110, opened 2016-11-10
Comment on return types

kvigen:
Add a comment on the return types in the Go interfaces. This is both to match the JS behavior and to make it more clear which errors the caller should check for.

### DIFF
--- a/samples/gen-go-deprecated/server/interface.go
+++ b/samples/gen-go-deprecated/server/interface.go
@@ -11,6 +11,12 @@ import (
 // Controller defines the interface for the swagger-test service.
 type Controller interface {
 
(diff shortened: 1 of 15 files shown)


samples/gen-go-deprecated/server/interface.go
```diff
@@ -11,6 +11,12 @@ import (
 // Controller defines the interface for the swagger-test service.
 type Controller interface {
 
-	// GetBook makes a GET request to /books/{id}.
+	// GetBook makes a GET request to /books/{id}
+
+	// 200:
```
rgarcia: maybe this should read `200: nil`?

also kind of weird that this method doesn't return a book...

PR #111, opened 2016-11-10
Increase duration error margin in test

kvigen:
To handle errors like:
Max difference between 2016-11-10 15:07:43.686317769 -0800 PST and
2016-11-10 15:07:43.676000766 -0800 PST allowed is 10ms, but difference
was 10.317003ms


rgarcia:
lgtm

PR #112, opened 2016-11-11
Remove empty space in genjs

cgclever:
This showed up bright red in my git diff


kvigen:
can you run `make test` to generate the new sample JS code. otherwise lgtm!

kvigen:
thanks, lgtm

PR #114, opened 2016-11-16
Convert nil arrays to empty arrays

kvigen:
Before this change if a wag app returned
```
var result []models.Type{}
```
if would get serialized into "null" by the json marshaler because it has
a nil type. Since we don't think the nil slice type is very useful, we
now convert nil slices to empty slices on output.


rgarcia:
lgtm

PR #115, opened 2016-11-17
Allow setting logger on WagClient

mohit:
This allows for:

* the consumer of the client code to set the logger
* Sets the logger `source` to `ServiceName` rahter than `wag`.


mohit:
tested by running `make generate` for `init-service` and using the client code in `ark`

kvigen:
lgtm!

PR #116, opened 2016-11-18
Clean up input parameter handling

kvigen:
Ideally we would have done this before v1, but I think the backwards incompatible changes won't break any current consumers, so we should sneak it in. This change does a couple things:

- It distinguishes between nil and empty input parameters where appropriate. Not doing this is causing a bug in the app-service.
- It documents the input parameter behavior in detail
- It tweaks the handling of headers since we can't distinguish between nil and empty in that case
- It adds a few tests

I'm not a huge fan of distinguishing between empty and nil values in some cases, but not others, but right now I think it's better than the alternative.

We should verify that the JS client also follow all these conventions


kvigen:
cc @rgarcia

cozmo:
FYI @evaninja @ahsojar

cozmo:
@kvigen This PR is a little hard to follow line by line but overall the changes to the generated code and the code generation look reasonable so if you're comfortable with it I am too.

carolinemodic:
Will this PR mean that you can no longer call getApps with no input parameters? If so district-authorizations-launcher will break.

kvigen:
@carolinemodic you can still call with input parameters. The difference is that now the server can distinguish between you passing in the empty string and nil (left out of the query params). In the case where it isn't specified at all we don't filter by the field, in the case where it's the empty string we will match against that.

PR #117, opened 2016-11-19
use public npm package for discovery

mohit:
@clever/discovery -> clever-discovery


mohit:
how do I regenerate the samples?

kvigen:
`make test`

kvigen:
other than that lgtm

mohit:
@kvigen I think this should be good now.

PR #119, opened 2016-11-28
Handle headers correctly

cozmo:
I _think_ header handling was pretty broken before this change. 

1. Using `r.Header["header_name"]` to get the value of a header is...risky since the case of the keys in that map is unclear. At the very least the first character will be capitalized so if you use `header_name` as the swagger param it won't work (since in the map it'll be `Header_name`). Switch (back) to using `r.Header.Get()`

TBD if/how we want to add some sort sort of test to ensure this doesn't happen going forward.

### DIFF
--- a/samples/gen-go-deprecated/server/handlers.go
+++ b/samples/gen-go-deprecated/server/handlers.go
@@ -147,8 +147,8 @@ func newHealthInput(r *http.Request) (*models.HealthInput, error) {
 	}
 
 	if len(sectionStrs) > 0 {
-		sectionStr := sectionStrs[0]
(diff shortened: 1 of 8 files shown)


kvigen:
As we discussed, headers returns `[]string` so this is only broke for non-canonical headers.

kvigen:
I can't find anything beyond a trivial test for headers. It shouldn't be too hard to add one in here: https://github.com/Clever/wag/blob/master/test/server_test.go. I'm also happy adding this later because this makes me want to do a more thorough audit of our testing.

kvigen:
lgtm!

PR #121, opened 2016-11-30
Tag spans

kvigen:
Adding the common tags documented here:
https://github.com/opentracing/opentracing.io/blob/95b966bd6a6b2cf0f231260e3e1fa6206ede2151/_docs/pages/api/data-conventions.md#component-identification

This also tries to associate a clearer operation name with the span. The way we do it is a bit hacky, but since it's fairly hidden behind an interface I think I'm okay with it. Will keep thinking about it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -181,6 +181,8 @@ func main() {
     * **wag**: All `wag`-generated clients take in a context object as the first argument.
       If your handler consumes a `wag`-generated client, then pass the context object to these client methods.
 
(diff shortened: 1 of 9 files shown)


kvigen:
should probably switch this to not use constant https://github.com/opentracing/opentracing-go/blob/master/ext/tags.go

cozmo:
LGTM - This was very easy to follow 👍

PR #122, opened 2016-12-02
Refactor code generation

kvigen:
This change will make it easier to support arbitrary single parameter methods.

For the most part it tries to move all the `if` and `printf` statements into templates so that they're easier to read. It also replaces things like `i.%s` with `{{.AccessString}}`

I do this for both the model and client generation. The server generation was fairly close to a good place already.

Did this separately so we could make sure the refactor kept the generated code the same (except for whitespace)

### DIFF
--- a/clients/go/gengo.go
+++ b/clients/go/gengo.go
@@ -303,9 +303,9 @@ func paramToTemplate(s *spec.Swagger, param *spec.Parameter, op *spec.Operation)
 		ErrorMessage: errorMessage(s, op),
 	}
 	if singleStringPathParameter {
-		t.Name = op.Parameters[0].Name
+		t.Name = param.Name
 		t.ToStringCode = singleParamName
-		t.AccessString = op.Parameters[0].Name
+		t.AccessString = param.Name
 	}
 	return t
 }
--- a/clients/go/gengo.go
+++ b/clients/go/gengo.go
@@ -281,6 +281,10 @@ func methodCode(s *spec.Swagger, op *spec.Operation, basePath, method, methodPat
 
 }
 
+// paramToTemplate converts a spec.Parameter into the a struct with the information needed to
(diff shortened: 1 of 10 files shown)


clients/go/gengo.go
```diff
@@ -281,79 +281,74 @@ func methodCode(s *spec.Swagger, op *spec.Operation, basePath, method, methodPat
 
 }
 
+func paramToTemplate(s *spec.Swagger, param *spec.Parameter, op *spec.Operation) paramTemplate {
```
cozmo: I think it would be ideal to have a comment explaining what this function does - I have no idea.

samples/gen-go/client/client.go
```diff
@@ -248,9 +257,12 @@ func (c *WagClient) CreateBook(ctx context.Context, i *models.Book) (*models.Boo
...
 		var err error
 		body, err = json.Marshal(i)
-
 		if err != nil {
-			return nil, err
+
+			if err != nil {
+				return nil, err
+			}
+
```
cozmo: We seem to be doing the same check twice here, which is logically sound but a little weird.

cozmo:
The generated code didn't change much so this in general seems reasonable. Don't fully understand all the changes so would love a comment or two but otherwise LGTM.

samples/gen-go/client/client.go
kvigen: https://github.com/Clever/wag/pull/122/commits/d6cd82e43d0e1b99cdcbfac62f07e2597f974676

clients/go/gengo.go
kvigen: https://github.com/Clever/wag/pull/122/commits/34a0a25b30c55764fab0ef47cac39d233bbfac89

PR #124, opened 2016-12-12
Update README.md

drhurd:
README.md now uses "tracing-middleware" npm package to describe client usage in node services.

### DIFF
--- a/README.md
+++ b/README.md
@@ -252,7 +252,7 @@ To utilize the `span` option above you need to pass an opentracing span into the
 example shows you how to setup an express app to track requests and any calls made via a wag client.
 
 ```bash
-npm install --save "tracing-middleware"
+npm install --save tracing-middleware
 ```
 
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -252,51 +252,26 @@ To utilize the `span` option above you need to pass an opentracing span into the
 example shows you how to setup an express app to track requests and any calls made via a wag client.
 
 ```bash
-npm install lightstep-tracer opentracing@0.11 --save # >=0.12 contains untested breaking changes to the API
+npm install --save "tracing-middleware"
```
rgarcia: nit: i don't think the quotes are necessary?

README.md
```diff
@@ -252,51 +252,26 @@ To utilize the `span` option above you need to pass an opentracing span into the
 example shows you how to setup an express app to track requests and any calls made via a wag client.
```
kvigen: i might also throw in that if you want to guarantee that a trace is sampled you can set it in the request.span object after the middleware is run like this:
```
req.span.setTag("sampling.priority", 1)
```

drhurd: bb6d29b

PR #129, opened 2016-12-15
Update to latest process-metrics

kvigen:
Couldn't get glide to not update a bunch of other things, but I'm okay with that.


glide.lock
```diff
@@ -145,73 +135,94 @@ imports:
...
 - name: gopkg.in/Clever/kayvee-go.v5
   version: 865aa5e4bbe8e274c69accd53f0b1e2ecc484cd8
+  subpackages:
+  - logger
+  - middleware
+- name: gopkg.in/Clever/kayvee-go.v6
+  version: fa278c9c7f83580b15c163c1efffe6fb4f2e3989
+  subpackages:
+  - logger
+  - router
```
xavi-: Ouch, wag requires 3 versions of kayvee =/

kvigen: yup...

PR #131, opened 2016-12-16
Error in client for empty path parameters

kvigen:
Because if the path parameters are empty then we can't construct the URL.

### DIFF
--- a/clients/js/genjs.go
+++ b/clients/js/genjs.go
@@ -243,10 +243,19 @@ var methodTmplStr = `
     const timeout = options.timeout || this.timeout;
     const span = options.span;
 
-    const headers = {};{{range $param := .HeaderParams}}
-    headers["{{$param.WagName}}"] = params.{{$param.JSName}};{{end}}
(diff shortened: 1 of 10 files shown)


kvigen:
cc @samfishman

clients/js/genjs.go
```diff
@@ -243,10 +243,19 @@ var methodTmplStr = `
     const timeout = options.timeout || this.timeout;
     const span = options.span;
 
-    const headers = {};{{range $param := .HeaderParams}}
-    headers["{{$param.WagName}}"] = params.{{$param.JSName}};{{end}}
-
-    const query = {};{{range $param := .QueryParams}}{{ if $param.Required }}
+    const headers = {};
+    {{- range $param := .PathParams}}
+    if (params.{{$param.JSName}} == "") {
```
cozmo: Could do this just as `if (!params.{{$param.JSName}}) {` which would catch the undefined case as well. 

If we don't do that we should use `===` I think.

PR #132, opened 2016-12-16
automatically set up lightstep

rgarcia:
Starting to look into updating our opentracing stuff, and starting with wag.


rgarcia:
@kvigen can you take another look? I modified this to do what we discussed in https://github.com/Clever/template-wag/pull/16

PR #133, opened 2016-12-30
Handle undefined callback with empty path params

kvigen:
Missed the empty callback case

### DIFF
--- a/clients/js/genjs.go
+++ b/clients/js/genjs.go
@@ -236,51 +236,6 @@ var methodTmplStr = `
       options = undefined;
     }
 
-    if (!options) {
-      options = {};
-    }
-
-    const timeout = options.timeout || this.timeout;
-    const span = options.span;
-
-    const headers = {};
-    {{- range $param := .PathParams}}
-    if (!params.{{$param.JSName}}) {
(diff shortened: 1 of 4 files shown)


clients/js/genjs.go
```diff
@@ -246,7 +246,9 @@ var methodTmplStr = `
     const headers = {};
     {{- range $param := .PathParams}}
     if (!params.{{$param.JSName}}) {
-      cb(new Error("{{$param.JSName}} must be non-empty because it's a path parameter"))
+      if (cb) {
+        cb(new Error("{{$param.JSName}} must be non-empty because it's a path parameter"));
+      }
```
cozmo: I think this isn't quite right - I think we want to call the callback if it exists, else we want to reject the promise. That's what the `resolver` and `rejector` functions are for below, but those are in the promise. Not sure exactly how we want to solve this but we may just want to wrap the whole function body with the promise and then we can use `resolve` and `reject` like we do below?

kvigen: derp. yeah, wrapping the whole function body makes sense

kvigen: @cozmo back to you. what do you think?

cozmo:
Yea this looks good to me.

PR #134, opened 2017-01-03
upgrade version of opentracing to 0.13.0

rgarcia:


cozmo:
If we're happy that this works then LGTM.

rgarcia:
@cozmo just updated this w/ a test that hits the tracing code

cozmo:
Seems reasonable.

PR #136, opened 2017-01-05
check glide.ymls for correct dependencies

rgarcia:
This adds a check that users have pulled in the correct versions of dependencies that wag requires to function properly.

### DIFF
--- a/main.go
+++ b/main.go
@@ -46,6 +46,12 @@ func main() {
 		log.Fatal("js-path is required")
 	}
 
+	if glideYMLFile, err := os.Open("glide.yml"); err == nil {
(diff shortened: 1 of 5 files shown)


main.go
```diff
@@ -46,6 +46,12 @@ func main() {
 		log.Fatal("js-path is required")
 	}
 
+	if glideYMLFile, err := os.Open("glide.yml"); err == nil {
```
kvigen: so we're not going to enforce glide?

rgarcia: Oh interesting, I didn't think about going that far with this check. I'd be fine starting without enforcing glide, and later adding it in if we find it necessary.

PR #140, opened 2017-01-11
Explicitly error on nested object definitions

kvigen:
definitions:
  Foo:
    type: object
    properties:
      bar:
        type: object
        properties:
          baz:
           type: string

Generates an error:
gen-go/models/foo.go:44: m.Bar.Validate undefined (type *FooBar has no
field or method Validate)

This appears to be a bug in the way go-swagger generates validation
code. Let's make it clearer to uses what's going on by explicitly
disallowing this.

### DIFF
--- a/validation/validation.go
+++ b/validation/validation.go
@@ -139,6 +139,8 @@ func validateDefinitions(definitions map[string]spec.Schema) error {
 	for name, def := range definitions {
 		for _, subDef := range def.Properties {
 			if len(subDef.Type) == 1 && subDef.Type[0] == "object" {
+				// We throw an error here because nested objects generate a compiler error in
+				// the go-swagger model code.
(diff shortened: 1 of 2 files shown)


validation/validation.go
```diff
@@ -135,6 +135,17 @@ func validateParams(path, method string, op *spec.Operation) error {
 	return nil
 }
 
+func validateDefinitions(definitions map[string]spec.Schema) error {
+	for name, def := range definitions {
+		for _, subDef := range def.Properties {
+			if len(subDef.Type) == 1 && subDef.Type[0] == "object" {
+				return fmt.Errorf("%s cannot have nested object types", name)
```
rgarcia: I'd add a comment here describing why we can't have these, ideally linking to a go-swagger issue if it exists

PR #141, opened 2017-01-18
Export enum values

samfishman:
Updates go-swagger to the earliest version that exports string enums.


rgarcia:
lgtm, looks like circleci is failing?

one thing that my old PR does is add consts for enum input parameters (we don't use go-swagger for these). Not sure if you need this

samfishman:
Looks like circleci isn't properly set up yet.

And yeah, I don't need the input enums yet so I left that out.

PR #145, opened 2017-02-07
Fix params with dashes

samfishman:
Dashes were included in variable names in the generated handler code for parameters that had dashes, which doesn't make syntastic sense.

### DIFF
--- a/server/genserver.go
+++ b/server/genserver.go
@@ -9,6 +9,7 @@ import (
 
 	"github.com/Clever/wag/swagger"
 	"github.com/Clever/wag/templates"
+	"github.com/Clever/wag/utils"
 )
 
 // Generate server package for a swagger spec.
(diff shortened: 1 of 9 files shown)


server/genserver.go
```diff
@@ -601,37 +602,37 @@ type paramTemplate struct {
 
 var paramTemplateStr = `
 	{{if eq .ParamType "query" -}}
-		{{.ParamName}}Strs := r.URL.Query()["{{.ParamName}}"]
+		{{.VarName}}Strs := r.URL.Query()["{{.ParamName}}"]
```
cozmo: Why are we switching all of these to `.VarName`?

samfishman: `.VarName` is the camel-cased version of the parameter (so has dashes removed), whereas `.ParamName` is the literal name of the parameter.

PR #146, opened 2017-02-13
Paging server side

samfishman:
Adds scaffolding to the server-side so that route controllers can declare what the next page ID is – if paging is configured, controllers get an extra parameter to return (between the return value and error) that they can either pass the zero value to (implies no pages remaining) or a page ID. If a nonzero value is returned, the `X-Next-Page-Path` header is set to the path of the next page, which the autogenerated handler constructs.

Paging can be figured by adding an `x-paging` property to the operation config:
```yaml
  /books:
    get:
      operationId: getBooks
      description: Returns a list of books
      x-paging:
        pageParameter: startingAfter
```


swagger/operation.go
```diff
@@ -118,6 +151,26 @@ func SuccessType(s *spec.Swagger, op *spec.Operation) *string {
...
+	pagingParamName, ok := pagingConfig["pageParameter"].(string)
+	if !ok {
+		return spec.Parameter{}, false
+	}
+	for _, p := range op.Parameters {
+		if p.Name == pagingParamName {
+			return p, true
+		}
+	}
+	return spec.Parameter{}, false
```
kvigen: should we panic here? this means our validation screwed up, right?

swagger/operation.go
```diff
@@ -118,6 +151,26 @@ func SuccessType(s *spec.Swagger, op *spec.Operation) *string {
...
 
+// PagingParam returns the parameter that specifies the page ID for this
+// operation, if paging is configured. If paging is not configured, the second
+// return value is `false`.
+func PagingParam(op *spec.Operation) (spec.Parameter, bool) {
+	pagingConfig, ok := op.Extensions["x-paging"].(map[string]interface{})
+	if !ok {
+		return spec.Parameter{}, false
+	}
+	pagingParamName, ok := pagingConfig["pageParameter"].(string)
```
kvigen: should we just call this `parameter`? given that paging is a bit redundant?

swagger/operation.go
```diff
@@ -118,6 +151,26 @@ func SuccessType(s *spec.Swagger, op *spec.Operation) *string {
...
+	if !ok {
+		return spec.Parameter{}, false
```
kvigen: this case should be caught in validation right? panic instead?

server/genserver.go
```diff
@@ -397,9 +418,9 @@ func statusCodeFor{{.Op}}(obj interface{}) int {
 func (h handler) {{.Op}}Handler(ctx context.Context, w http.ResponseWriter, r *http.Request) {
 {{if .HasParams}}
 	{{if .SingleStringPathParameter}}
-		{{.SingleStringPathParameterVarName}}, err := new{{.Op}}Input(r)
+		{{.InputVarName}}, err := new{{.Op}}Input(r)
 	{{else}}
-		input, err := new{{.Op}}Input(r)
+		{{.InputVarName}}, err := new{{.Op}}Input(r)
```
kvigen: should we get rid of this if statement now?

server/genserver.go
```diff
@@ -460,6 +481,24 @@ func (h handler) {{.Op}}Handler(ctx context.Context, w http.ResponseWriter, r *h
 		return
 	}
 
+	{{if .HasPaging}}
+	if !swag.IsZero(nextPageID) {
+		{{if .SingleStringPathParameter}}
+		{{.InputVarName}} = nextPageID
+		w.Header().Set("X-Next-Page-Path", models.{{.Op}}InputPath({{.InputVarName}})
```
kvigen: indent this so the template `if` logic is easier to follow? gofmt will make it look good again

kvigen: indent this so the template if logic is easier to follow? gofmt will make it look good again

swagger/operation.go
```diff
@@ -28,12 +38,35 @@ func Interface(s *spec.Swagger, op *spec.Operation) string {
 		input = fmt.Sprintf("i *models.%sInput", capOpID)
 	}
 
-	successType := SuccessType(s, op)
-	if successType == nil {
-		return fmt.Sprintf("%s(ctx context.Context, %s) error", capOpID, input)
+	includeNames := false
+	returnTypes := []string{}
+	returnTypesWithNames := []string{}
```
kvigen: are we getting much value from return types with names? makes things a bit bit messier, and given that i never use them in actual go code i write, i'm curious if think they're worth it. This is something consumers would add to their controller if they really wanted them, right?

test/sample_server.go
```diff
@@ -42,15 +60,22 @@ func (c *ControllerImpl) GetBookByID2(ctx context.Context, id string) (*models.B
...
 		return nil, models.Error{}
 	}
 	return book, nil
 }
 
 // CreateBook creates a book.
 func (c *ControllerImpl) CreateBook(ctx context.Context, input *models.Book) (*models.Book, error) {
+	if int(input.ID) >= len(c.books) {
+		extension := make([]*models.Book, int(input.ID)-len(c.books)+1)
+		c.books = append(c.books, extension...)
```
kvigen: hrm, this is weird. should we just make the books stuff a map instead of trying to force it into an array?

server/genserver.go
kvigen: it looks like we don't have anything that tests this path?

test/server_test.go
```diff
@@ -48,8 +49,42 @@ func TestBasicEndToEnd(t *testing.T) {
...
+	assert.Equal(t, "", resp.Header.Get("X-Next-Page-Path"))
+
+	controller.pageSize = 1
+	_, err = c.CreateBook(context.Background(), &models.Book{
+		ID: int64(2), Name: "Second",
+	})
+	require.NoError(t, err)
+
+	// Now that there is a second page, X-Next-Page-Path should be set
+	resp, err = http.Get(fmt.Sprintf("%s/v1/books", s.URL))
```
kvigen: worth having the test set a query parameter that's not related to paging to make sure the URL gets set up correctly?

models/genmodels.go
```diff
@@ -202,6 +212,84 @@ var validationStr = `
 	{{- end }}
 `
 
+func printInputSerializer(g *swagger.Generator, op *spec.Operation, basePath, methodPath string) error {
```
kvigen: 👍  to moving this to the model

server/genserver.go
```diff
@@ -460,6 +481,24 @@ func (h handler) {{.Op}}Handler(ctx context.Context, w http.ResponseWriter, r *h
 		return
 	}
 
+	{{if .HasPaging}}
+	if !swag.IsZero(nextPageID) {
+		{{if .SingleStringPathParameter}}
+		{{.InputVarName}} = nextPageID
```
kvigen: can this never be a pointer?

swagger/operation.go
samfishman: I think it makes it clearer what each return value is for, but maybe it's just as clear if we just include it in the docstring in the interface.

server/genserver.go
samfishman: Yeah, since this is in the `SingleStringPathParameter` branch (and single-string-path-parameter only happens for _required_ single-string parameter functions, making it not a pointer).

samfishman: I fixed the indentation to make this `if` block more clear.

test/sample_server.go
samfishman: I forced them into an array so that I could get consistent ordering on output, but maybe I could just keep track of the max ID and loop through all possible IDs on output. Which do you think is better?

server/genserver.go
samfishman: Hmm I think maybe we should just disallow paging with `SingleStringPathParameter` since it's pretty weird to have the page parameter in the path. What do you think?

swagger/operation.go
samfishman: 43b91eb

swagger/operation.go
samfishman: 43b91eb

swagger/operation.go
samfishman: 43b91eb

server/genserver.go
samfishman: 43b91eb

samfishman: derp

server/genserver.go
samfishman: 43b91eb

swagger/operation.go
kvigen: i don't have a strong feeling. though maybe we should be consistent whatever way we go?

server/genserver.go
kvigen: i like that. worst case we add support for it back later

test/sample_server.go
kvigen: i think that would be a bit clearer than this, but i don't have a strong feeling. Either way might be worth a comment

swagger/operation.go
samfishman: If consistency is the most important thing here, it probably makes sense to just do it in a comment.

test/sample_server.go
samfishman: 4e84bc3

swagger/operation.go
samfishman: 4e84bc3

server/genserver.go
samfishman: ecf9ae2 (also updated validation PR)

samfishman:
@kvigen I think everything should be addressed now, let me know if it's ready to go.

PR #150, opened 2017-03-16
Fix typo in error message

peggyl:


kvigen:
`make test` should regenerate the test code

PR #151, opened 2017-03-27
Update to kayvee v6

meliaj:
Updates wag to use kayvee v6, and look for a kvconfig.yml file for log routing.


circle.yml
```diff
@@ -20,7 +20,7 @@ test:
   - $HOME/ci-scripts/circleci/report-card $RC_DOCKER_USER $RC_DOCKER_PASS "$RC_DOCKER_EMAIL" $RC_GITHUB_TOKEN
 deployment:
   all:
-    branch: master
```
kvigen: reminder to remove this

README.md
```diff
@@ -97,17 +97,26 @@ The server interface defined in `gen-go/server/interface.go` has one method for
...
 ```go
-import "gopkg.in/Clever/kayvee-go.v5/logger"
+import "gopkg.in/Clever/kayvee-go.v6/logger"
 ...
 logger.FromContext(ctx).Info(...)
 ```
 
   You should use this logger for all logging within your controller implementation.
 
+#### Application Log Routing
```
kvigen: this repo is public, so i wonder if we should discuss log routing without making it clear it's not clever-specific.

should we also add this to: https://github.com/Clever/template-wag with an empty kv config? Doesn't necessarily need to be you who makes the change

meliaj: Added a note here https://github.com/Clever/wag/pull/151/commits/48b8300ea26659e1b5c48c94db4dd10d5c867722

Also going to add empty kvconfig.yml to template-wag

PR #154, opened 2017-05-02
Use kayveelogger for setlogger

mohit:
Clients generated with the latest version of wag does not support `SetLogger` properly since `kayvee.New()` now returns `kayvee.KayveeLogger` interface.

`client.SetLogger(*kayvee.Logger)` was using a struct as opposed to the interface `kayvee.KayveeLogger`. Fix and add test to ensure the right `version`/`interface` is being passed to the client.

### DIFF
--- a/clients/go/gengo.go
+++ b/clients/go/gengo.go
@@ -117,7 +117,7 @@ func (c *WagClient) SetCircuitBreakerDebug(b bool) {
 	c.circuitDoer.debug = b
 }
 
-// SetLogge  allows for setting a custom logger
+// SetLogger allows for setting a custom logger
 func (c *WagClient) SetLogger(logger logger.KayveeLogger) {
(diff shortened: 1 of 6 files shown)


clients/go/gengo.go
```diff
@@ -117,8 +117,8 @@ func (c *WagClient) SetCircuitBreakerDebug(b bool) {
 	c.circuitDoer.debug = b
 }
 
-// SetLogger allows for setting a custom logger
-func (c *WagClient) SetLogger(logger *logger.Logger) {
+// SetLogge  allows for setting a custom logger
```
rgarcia: SetLogger*

mohit: derp https://github.com/Clever/wag/pull/154/commits/fac930d2fddde7cd8cb0197d27325c1119f87689

PR #155, opened 2017-05-04
Support paging on put

kshen0:
Adds support for paging on PUTs. In general requests that need paging should be on GETs, but in some cases endpoints may need to handle request bodies and also support paging.


validation/validation.go
```diff
@@ -149,7 +149,9 @@ func validatePaging(s *spec.Swagger, path, method string, op *spec.Operation) er
 		return nil
 	}
 
-	if strings.ToUpper(method) != "GET" {
+	upperMethod := strings.ToUpper(method)
+	// Not a technical limitation, but semantic
```
rgarcia: nit: I don't understand what this means

kshen0: clarified in 9c61c2d

PR #160, opened 2017-07-20
Add client side Kayvee error logs for golang

mohit:
Add a structured log in case the backend fails to respond with a successful response after the configured retries.


clients/go/gengo.go
```diff
@@ -381,10 +381,15 @@ func (c *WagClient) do%sRequest(ctx context.Context, req *http.Request, headers
 	}
 	resp, err := c.requestDoer.Do(client, req)
 	if err != nil {
+		c.logger.ErrorD("client-request-finished", map[string]string{
```
kvigen: do we want to put error in the title so it's clear that it's this is logged on errors.

Also, when we get a 50x, err will be nil, but resp.StatusCode will be >= 500. Do we want to track those too?

mohit: I'm hoping we can in the future add this in non error conditions too, but I think we could add the `error` in the signalfx series at the matcher level. 

Yes. Thanks for pinging - it is a bug: 

TIL: ```A non-2xx status code doesn't cause an
  // error.```
https://golang.org/src/net/http/client.go?s=16536:16588#L463

mohit: @kvigen here is an updated attempt: https://github.com/Clever/wag/pull/160/commits/205cd47ba9b2f625e2ff943f3f7c7a4154365db7

clients/js/genjs.go
```diff
@@ -313,6 +322,11 @@ var methodTmplStr = `
       async.whilst(
         () => requestOptions.uri !== "",
         cbW => {
+	  this.logger.errorD("client-request-finished", {
+		"backend": "{{.ServiceName}}",
+		"message": arguments.length > 0 ? arguments[0] : "",
+		"status_code": 0
```
kvigen: always status code of zero?

mohit: that was just a test implementation. Actual implementation after talking with @cozmo https://github.com/Clever/wag/pull/160/commits/4c807adaa667887a308f23bae8f6b29bb017a416

clients/js/genjs.go
```diff
@@ -338,10 +374,12 @@ var methodTmplStr = `
 
           switch (response.statusCode) {
             {{ range $response := .Responses }}case {{ $response.StatusCode }}:{{if $response.IsError }}
+			  var err = new Errors.{{ $response.Name }}(body || {});
+			  responseLog(this.logger, response, err);
```
mohit: @cozmo seems like I'm having issues with `this` here. Would love to walk through with you tomorrow?

samples/gen-js/index.js
```diff
@@ -68,6 +69,27 @@ const noRetryPolicy = {
 };
 
 /**
+ * Request status log is used to 
+ * to output the status of a request returned 
+ * by the client.
+ */
+function responseLog(logger, response, err) {
```
mohit: this is the final rendered function responseLog
 @cozmo

samples/gen-js/index.js
```diff
@@ -188,6 +217,7 @@ class SwaggerTest {
 
       const retryPolicy = options.retryPolicy || this.retryPolicy || singleRetryPolicy;
       const backoffs = retryPolicy.backoffs();
+      const logger = this.logger;
   
       let retries = 0;
       (function requestOnce() {
```
mohit: calling responseLog @cozmo

mohit:
Tested the node client by connecting in success and failure cases with workflow-manager.

clients/go/gengo.go
```diff
@@ -380,11 +387,29 @@ func (c *WagClient) do%sRequest(ctx context.Context, req *http.Request, headers
 	    req = req.WithContext(ctx)
 	}
 	resp, err := c.requestDoer.Do(client, req)
+	retCode := 0
+	if resp != nil {
+	  retCode = resp.StatusCode
+	}
+
+	// log all client failures and non-successful HT
```
mohit: @kvigen added `method` and `uri`

PR #161, opened 2017-07-25
add circuit breaking to JS client

rgarcia:
We have circuit breaking in the Go client, but not in the JS client. This PR adds circuit breaking to the js client.

I tried to make this as similar as possible to how the go client does it:
- It exposes the same settings (w/ same defaults) as the Go client circuit breaker
- It periodically logs circuit state

### DIFF
--- a/clients/js/genjs.go
+++ b/clients/js/genjs.go
@@ -188,8 +188,7 @@ const defaultCircuitOptions = {
   requestVolumeThreshold: 20,
   sleepWindow:            5000,
   errorPercentThreshold:  90,
-  logIntervalMs:          30000,
-  logger: () => { /* TODO remove this and use kayvee logger from mohit's PR */ },
+  logIntervalMs:          30000
 };
 
 /**
@@ -270,7 +269,7 @@ class {{.ClassName}} {
       context(this).
       build();
 
-    setInterval(() => this._logCircuitState(circuitOptions.logger), circuitOptions.logIntervalMs);
+    setInterval(() => this._logCircuitState(), circuitOptions.logIntervalMs);
   }
 
   _hystrixCommandErrorHandler(err) {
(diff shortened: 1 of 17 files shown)


clients/js/genjs.go
```diff
@@ -152,6 +154,21 @@ const noRetryPolicy = {
...
+ */
+const defaultCircuitOptions = {
+  debug:                  true,
+  requestVolumeThreshold: 20,
+  maxConcurrentRequests:  100,
+  requestVolumeThreshold: 20,
+  sleepWindow:            5000,
+  errorPercentThreshold:  90,
+  logIntervalMs:          30000,
+  logger: () => { /* TODO remove this and use kayvee logger from mohit's PR */ },
```
rgarcia: cc @mohit :)

clients/js/genjs.go
```diff
@@ -174,6 +191,17 @@ class {{.ClassName}} {
    * in milliseconds. This can be overridden on a per-request basis.
    * @param {module:{{.ServiceName}}.RetryPolicies} [options.retryPolicy=RetryPolicies.Single] - The logic to
    * determine which requests to retry, as well as how many times to retry.
+   * @param {Object} [options.circuit] - Options for constructing the client's circuit breaker.
+   * @param {bool} [options.circuit.debug] - When set to true will not circuit break. Default: true.
```
cozmo: This doesn't seem like the most well named option? Does all it do is disable? Does it also log more?

rgarcia: I am passing this setting directly through to the `circuitBreakerForceClosed` setting, so maybe it could be named `forceClosed`?

clients/js/genjs.go
mohit: merged now: https://github.com/Clever/wag/pull/160

clients/js/genjs.go
```diff
@@ -226,6 +253,54 @@ class {{.ClassName}} {
...
+    // to avoid counting 4XXs as errors, only count an error if it comes from the request library
+    if (err._fromRequest === true) {
+      return err;
+    }
+    return false;
+  }
+
+  _hystrixCommandRun(method, args) {
+    return method.apply(this, args);
+  }
```
cozmo: Seems like a passthrough method - do we need this? Don't know a ton about hystrix.

rgarcia: The hystrix circuit breaker needs to be constructed with a method to run, which is tricky when our client class has many methods (one per API operation). I get around this with `_hystrixCommandRun`, which is effectively a proxy method to all the methods on the class.

PR #162, opened 2017-08-04
Require latest gomock

rgarcia:
gomock made breaking changes that cause build errors, e.g. https://circleci.com/gh/Clever/mpt-data-gator/142

### DIFF
--- a/glide.yaml
+++ b/glide.yaml
@@ -13,6 +13,8 @@ import:
   subpackages:
   - generator
 - package: github.com/golang/mock
+  # if you change this, please update validation/glide.go
+  version: 13f360950a79f5864a972c786a10a50e44b69541
   subpackages:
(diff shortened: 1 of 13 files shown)


glide.yaml
```diff
@@ -13,6 +13,8 @@ import:
   subpackages:
   - generator
 - package: github.com/golang/mock
+  # if you change this, please update validation/glide.go
+  version: 13f360950a79f5864a972c786a10a50e44b69541
```
mohit: How does this interact with an app using master in a wag server?

rgarcia: it'll ask them to switch to this version exactly

rgarcia:
@mohit pushed da6342b which will make sure the version of mockgen used during build / generate / test is the same as well

PR #163, opened 2017-10-18
update to latest go-swagger

rgarcia:
Interested in this fix: https://github.com/go-swagger/go-swagger/pull/1189 which gives people the option to add `x-omitempty` to array fields. This causes the json struct tag for the array to include `omitempty`.

This is useful, for example, for optional array fields like `Retry` / `Choice` in state machines. Currently these fields marshal to `null` if not provided, but omission is preferred.


rgarcia:
@mohit yes, also the changes in generated go clients are pretty trivial and safe

PR #164, opened 2017-10-26
Add stricter check and clearer error for peer dependency missing

nathanleiby:
https://clever.atlassian.net/browse/INFRA-2622

Now checks if both glide.yaml and glide.lock are valid.
Also adds more detailed error message.

**Before:**

image

**After:**

image

Does not yet roll up all errors at once.

### DIFF
--- a/main.go
+++ b/main.go
@@ -46,9 +46,21 @@ func main() {
 		log.Fatal("js-path is required")
 	}
 
-	if glideYMLFile, err := os.Open("glide.yaml"); err == nil {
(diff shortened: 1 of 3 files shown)


nathanleiby:
cc @alsmola

nathanleiby:
Added a rollup for peer dependency errors, so the new UX is:

### if `glide.yaml` of date

1- missing deps in glide yaml

### if `glide.yaml` up to date, but `glide.lock` out of date

1- wrong versions in glide lock

nathanleiby:
Need to fix tests, will do so momentarily

PR #166, opened 2017-12-06
WIP caching

rgarcia:
- adds a `x-caching` extension that you can configure endpoints with. Only supports adding `max-age` headers on responses

- adds caching support go clients via https://github.com/gregjones/httpcache

not done:

- `gregjones/httpcache` has an in memory cache we could use, but it doesn't limit the size of the cache in any way. `wag` should probably wrap this to have a limit

- `wag` should report a counter of cache hits/misses

- caching in the js client 😨


rgarcia:
@kvigen @samfishman this morning @Sayan- and I added cache hit/miss logging to this.

I think we can use https://github.com/die-net/lrucache as the cache implementation, since this will give us byte-size limiting and also some nice-to-have LRU eviction logic.

I think the next steps from here:
(- optional, if @samfishman wants to have a go at the JS client, add JS client caching)
- Get this reviewed/merged to roll a new wag release. None of the new code gets triggered unless a client calls `SetCache(...)` or a server adds `x-caching`, so this step should be safe.
- Change a client/server pair to use this, most likely hall-monitor + app-service

clients/go/gengo.go
```diff
@@ -111,6 +113,47 @@ func NewFromDiscovery() (*WagClient, error) {
...
+func newCacheHitCounter(cache httpcache.Cache, basePath string, l logger.KayveeLogger) *cacheHitCounter {
+	chc := &cacheHitCounter{Cache: cache, basePath: basePath}
+	go chc.log(l)
+	return chc
+}
+
+type cacheHitCounter struct {
+	httpcache.Cache
+	hits     int64
+	misses   int64
```
kvigen: probably not worth it yet, but i could see us wanting to know this on a per-endpoint basis

README.md
```diff
@@ -191,6 +191,20 @@ COPY bin/my-wag-service /usr/bin/my-wag-service
     that exposes `map`, `forEach`, and `toArray` functions to iterate over the
     results, again requesting new pages as needed.
 
+### Caching
+  * Wag can add `max-age` cache-control headers on respones if you use the `x-caching`
```
kvigen: nit: mention that max-age is in seconds

samples/swagger.yml
```diff
@@ -94,6 +94,28 @@ paths:
           schema:
             $ref: "#/definitions/Error"
 
+  /bookscached/{id}:
+    get:
+      operationId: getBookByIDCached
+      description: Retrieve a book
+      x-caching:
+        max-age: 3
```
kvigen: does it make sense to make this 1 so the tests run more quickly, or does that make the tests unstable?

kvigen (commented):
looks good!

have we vetted https://github.com/die-net/lrucache carefully? it doesn't have a ton of github stars and lru cache performance can be tricky. an alternative might be: https://github.com/karlseguin/ccache, but i don't know how difficult that will be to integrate with httpcache

also, thinking more about app-service + HM, i'm realizing that we have a challenge that when someone makes a change in HM they expect to see the change show up immediately. Not sure if there this is going to cause problems with the endpoints we're thinking about, or whether we should think about that case too

PR #168, opened 2018-01-12
Improve parameter missing error messages

nathanleiby:
Improves missing parameter error messages.

Previously

1. didn't say what type of parameter (header, path, query, body)
2. didn't say name of the missing parameter

- f2240bd clarifies the missing parameter error messages
- c3a993d makes a body field required, so we can see that in our sample


nathanleiby:
@alsmola I actually don't know, my guess was that a previous commit didn't update the samples after making a change

nathanleiby:
Lemme make sure all my deps are correct locally... maybe it's me 😨

nathanleiby:
Looks like my dependencies were out of date locally. Thanks for the nudge! I've reverted that commit (c3b9740f0787ae4bd8cf1414badb0e613c0f288b). Running another `make generate` locally shows no changes now.

alsmola:
:1:

kvigen:
lgtm too!

PR #170, opened 2018-01-20
Change default timeouts

kvigen:
Part of this is followup for a few of the recent ALB flares. It seems like our default timeout for Node requests should be 5 seconds, and I think we should try decreasing the Go timeout to 5 seconds too.

This isn't a full solution for Go. Since we reuse the context with the timeout across retries so the retries will fail immediately (I believe - haven't verified). Will follow up with a PR to address that.

### DIFF
--- a/VERSION
+++ b/VERSION
@@ -1,2 +1,3 @@
-v1.5.5
-- Log more frequently to match metrics rolling window
+v1.6.0
+- Added default 5 second timeout for Node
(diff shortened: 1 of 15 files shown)


kvigen:
cc @rgarcia @cozmo

kvigen:
Yeah, a minor version bump seems like a good idea.

PR #175, opened 2018-03-29
auto-generate a db package + dynamo implementation + tests

rgarcia:
Adds an extension to schema definitions that, when present, generate a `db` package. E.g.:

swagger.yml snippet:
```yaml
  ThingWithDateRange:
    x-db:
      AllowOverwrites: true
      DynamoDB:
        KeySchema:
          - AttributeName: name
            KeyType: HASH
          - AttributeName: date
            KeyType: RANGE
    type: object
    properties:
      name:
        type: string
      date:
        type: string
        format: date-time
```

db interface:
```go
// Interface for interacting with the swagger-test database.
type Interface interface {
	// SaveThingWithDateRange saves a ThingWithDateRange to the database.
	SaveThingWithDateRange(ctx context.Context, m models.ThingWithDateRange) error
	// GetThingWithDateRange retrieves a ThingWithDateRange from the database.
	GetThingWithDateRange(ctx context.Context, name string, date strfmt.DateTime) (*models.ThingWithDateRange, error)
	// DeleteThingWithDateRange deletes a ThingWithDateRange from the database.
	DeleteThingWithDateRange(ctx context.Context, name string, date strfmt.DateTime) error
}
```

tests for interface implementations: https://github.com/Clever/wag/blob/generate-db/samples/gen-go-db/server/db/tests/tests.go

dynamodb implementation:
- https://github.com/Clever/wag/blob/generate-db/samples/gen-go-db/server/db/dynamodb/dynamodb.go
- https://github.com/Clever/wag/blob/generate-db/samples/gen-go-db/server/db/dynamodb/thingwithdaterange.go


README.md
```diff
@@ -203,7 +203,35 @@ COPY bin/my-wag-service /usr/bin/my-wag-service
...
+    * context.TODO() - use this when you haven't been passed a context from a caller yet, but you expect the caller to send you one at some point.
+
+
+### DynamoDB Codegen
+
+  * Wag can auto-generate server code to save models to DynamoDB if you specify the `x-db` extension on a schema:
+  ```yaml
+definitions:
+  Thing:
+    x-db:
```
nathanleiby: these docs appear inconsistent with usage in `workflow-manager` https://github.com/Clever/workflow-manager/pull/133/files#diff-65058f891490afbb80c8704444e0af9bR392

samples/gen-go-db/server/handlers.go
```diff
@@ -0,0 +1,123 @@
...
+	"strconv"
+
+	"github.com/Clever/wag/samples/gen-go-db/models"
+	"github.com/go-errors/errors"
+	"github.com/go-openapi/strfmt"
+	"github.com/go-openapi/swag"
+	"github.com/gorilla/mux"
+	"gopkg.in/Clever/kayvee-go.v6/logger"
+)
+
```
nathanleiby: I'm familiar with this syntax.

In the other file you have a comment `// These imports may not be used depending on the input parameters"`

=> Still not sure I understand it.
=> Probably want to include a comment in all places you do this, if relevant

rgarcia:
@nathanleiby thanks for reviewing this... there are a lot files changed so here is my attempt to highlight what has changed / what should be reviewed:
- README.md documenting `x-db` extension
- Anything in the `samples/gen-go-db/server/db/` subdirectory. This is the new, auto-generated code from the new example `samples/db.yml` swagger file. Everything else in samples/gen-go-db can be ignored
- `server/gendb/` contains the codegen code

nathanleiby:
Thanks for the guidance. I've got a few notes from my first pass which I'll add now, and go through things again shortly.

README.md
```diff
@@ -203,7 +203,35 @@ COPY bin/my-wag-service /usr/bin/my-wag-service
 
   * If you don't have a context to pass to a Wag function you have two options
     * context.Background() - use this when this is the creator of the request chain, like a test or a top-level service.
-    * context.TODO() - use this when you haven't been passed a context from a caller yet, but you expect the caller to send you one at some point.  
+    * context.TODO() - use this when you haven't been passed a context from a caller yet, but you expect the caller to send you one at some point.
+
+
+### DynamoDB Codegen
+
+  * Wag can auto-generate server code to save models to DynamoDB if you specify the `x-db` extension on a schema:
```
nathanleiby: Do we imagine other dbs will be supported? Shoul we namespace to DynamoDB (`x-dynamodb`?)

README.md
```diff
@@ -203,7 +203,35 @@ COPY bin/my-wag-service /usr/bin/my-wag-service
...
+            KeyType: RANGE
+    type: object
+    properties:
+      name:
+        type: string
+      version:
+        type: integer
+  ```
+  The above will generate a `db` package with code to load/persist `Thing` objects from/to DynamoDB.
+  * `AllowOverwrites` specifies whether the auto-generated `Save` method should succeed if the object already exists.
```
nathanleiby: Maybe we could eliminate the need for this config by having more methods on the DB interface that have specific behaviors.

- Create (fails if already exists)
- Update (fails if DNE yet)
- Upsert / CreateOrUpdate / etc (creates or updates as needed)

README.md
```diff
@@ -203,7 +203,35 @@ COPY bin/my-wag-service /usr/bin/my-wag-service
...
+  * `DynamoDB` specifies the configuration for a DyanmoDB table for the schema.
+     It follows the format of the [`AWS::DynamoDB::Table`](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-dynamodb-table.html) CloudFormation resource.
+     Currently it supports a subset of the configuration allowed there.
```
nathanleiby: Might be nice to link a real working example from here (workflow-manager)

nathanleiby: Because it would show invoking the model from the controller, rather than just the auto-generated code that doesn't invoke the DB methods yet.

README.md
nathanleiby: After reading further, I also wonder whether having `AllowOverwrites` as a bool is flexible enough. 

I could imagine at the application level there might be some times when overwriting is OK and other times where it's not.

README.md
nathanleiby: Another thing I'm curious to see in action is `CreateTables`... is it just useful for the unit tests? do we think expect our apps to generate tables at runtime or continue to have them created out of band? generating tables at runtime would be especially awesome in dev, but makes less sense in prod. Also configuration like autoscaling isn't captured here.

samples/gen-go-db/server/db/dynamodb/dynamodb-local.sh
```diff
@@ -0,0 +1,21 @@
+#!/usr/bin/env bash
```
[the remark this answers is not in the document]
rgarcia: Interesting... my thought was that really only the server should be accessing the db

README.md
rgarcia: My thinking here:
- yes, in the future it may support other db types
- `x-db` top level config should be independent of db-type
- `DynamoDB` top level config contains DynamoDB-specific stuff

README.md
rgarcia: For now I'm designing this mostly to target the `workflow-manager` use case. I worry about over-designing it with future / unknown use cases in mind, so I'd rather evolve the design in conjunction with real use cases. Right now the config is tailored for `WorkflowDefinition` objects, which cannot be overwritten, and I expect it to evolve as it expands to support other objects

README.md
rgarcia: 👍 will add link to wfm once this is merged and wfm is updated

Right now `CreateTables` is only used in test, and the only configuration supported is primary `KeySchema`. In the future I could see
- Supporting more config options that influence the codegen in the swagger spec: TTL, secondary indexes (these are things that wfm requires)
- Figuring out some strategy for creating tables per-environment. I lean towards some mechanism where catapult includes the dynamo tables in the cloudformation template for an app. How that gets from the swagger spec to the template, I'm not sure...

rgarcia: Also in the future I could see catapult/ark owning setting autoscaling config. Here I lean towards something not configured in code, but rather changed on the fly via `ark`, e.g. `ark scale`

nathanleiby:
💥

PR #177, opened 2018-04-11
gendb: support querying for multiple items if table has range key

rgarcia:
e.g. this swagger def:
```yaml
  Thing:
    x-db:
      AllowOverwrites: false
      DynamoDB:
        KeySchema:
          - AttributeName: name
            KeyType: HASH
          - AttributeName: version
            KeyType: RANGE
```

now has an additional method to query for multiple `Thing`s:

```go
// GetThingsByNameAndVersion retrieves a list of Things from the database.
GetThingsByNameAndVersion(ctx context.Context, input GetThingsByNameAndVersionInput) ([]models.Thing, error)
```

```go
// GetThingsByNameAndVersionInput is the query input to GetThingsByNameAndVersion.
type GetThingsByNameAndVersionInput struct {
	Name                  string
	VersionStartingAt  *int64
	Descending            bool
	DisableConsistentRead bool
}
```


server/gendb/interface.go.tmpl
```diff
@@ -36,8 +41,34 @@ type Interface interface {
...
+{{- $indexName := indexName $xdbConfig.DynamoDB.KeySchema }}
+{{- $hashKey := index $xdbConfig.DynamoDB.KeySchema 0 }}
+{{- $hashKeySchema := index $xdbConfig.Schema.Properties $hashKey.AttributeName }}
+{{- if indexHasRangeKey $xdbConfig.DynamoDB.KeySchema }}
+{{- $rangeKey := index $xdbConfig.DynamoDB.KeySchema 1 }}
+{{- $rangeKeySchema := index $xdbConfig.Schema.Properties $rangeKey.AttributeName }}
+// Get{{ $modelName }}sBy{{ $indexName }}Input is the query input to Get{{ $modelName }}sBy{{ $indexName }}.
+type Get{{ $modelName }}sBy{{ $indexName }}Input struct {
+	{{ pascalize $hashKey.AttributeName }} {{ goType $hashKeySchema }}
+        {{ pascalize $rangeKey.AttributeName }}StartingAfter *{{ goType $rangeKeySchema }}
```
mohit: Am I understanding correctly that this is not always `StartingAfter`? Based on the `Descending` it would be before or after? Maybe `StartingFrom` is a better way to say this?

rgarcia: It'll always be `After`. `Descending` controls the ordering of the results returned, not the traversal of the index

rgarcia: I do like the name `From` better though, since the constraint is `>=`. Will change it to that

rgarcia: renamed to `...StartingAt`: 183c484

PR #184, opened 2018-05-23
Generate typescript typings file for javascript client

nathanleiby:
This is an MVP of auto-generating types for the javascript client.

So far, it just has common Wag client types, so we'll have strict checks when you run `new Client(<args>)`. It also has the method names with `any` interfaces.

It basically works in the current state and could be merged, I think.

--

Main work remaining to be done:

- generate types for all the `definitions` from the swagger.yml file, so we have the input/output types for methods
- update the `interface {{$ServiceName}}` methods to use these input/output types
- add unit tests to verify the typings


nathanleiby:
Fixed by #217

PR #186, opened 2018-07-18
allow optional keepAlive for node clients

mohit:
By default `request` and the node HTTP Agent do not enable `keepAlive`. Setting `forever: true` enables `keepAlive` for node version > 0.12. For older versions the `request-forever` agent is used to provide custom request pools.

Setting this up in a way that does not change existing behavior but allows client consumers to opt-in to set keepAlive.

### DIFF
--- a/VERSION
+++ b/VERSION
@@ -1,2 +1,2 @@
-v1.9.0
-- Allow setting http Transport
+v1.10.0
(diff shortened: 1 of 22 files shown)


cozmo:
LGTM but can you squash and merge your commits? Also you might need to bump the wag version?

PR #188, opened 2018-08-07
retry even if there is an error

nikhilpandit:


peternga (approved):

mohit:
we should carefully watch error rates and latencies in today's test, since this might have some implications there.

cozmo:
Looks like we didn't run `make generate` or whatever here. The sample code hasn't been updated.

cozmo (approved):

PR #193, opened 2018-10-29
Il 816 actually add span to header

peternga:
https://clever.atlassian.net/browse/IL-816

The `opentracing.inject` function wasn't actually injecting the span into the header -- we need to get the tracer before attempting to inject.

Tested this by modifying the app-service client code in oauth, inserting log messages in both the client library and in the app-service server. Verified the trace ID was being passed into the header and it was parsed on the server side.


rgarcia:
From those docs it seems like we should be getting the global tracers via `opentracing.globalTracer()` and not importing it like in this PR `const globalTracing = require("opentracing/lib/global_tracer");`

peternga:
Let me look into it a little deeper. What I got when I tried it earlier: `TypeError: opentracing.globalTracer is not a function`.

rgarcia:
@peternga just took a look, this looks great!

PR #196, opened 2018-11-27
support required fields in dynamo codegen

rgarcia:
Required fields in a swagger model definition cause struct fields to become pointers. This PR handles that case.


rgarcia:
@toddobryan-clever everything in `gen-go-db/` is autogenerated. Files in `gen-go-db/models/` are generated by a dependency, https://github.com/go-openapi. go-openapi makes required fields pointers so that validation can check for the existence or non-existence of the field

PR #201, opened 2019-04-10
SYNC-1014: dynamo paging

aastein:
https://clever.atlassian.net/browse/SYNC-1014

Adds a `GetThingsByKeysPage(ctx, input, fn)` for tables with range keys and tables with secondary indexes with range keys.


samples/gen-go-db/server/db/dynamodb/dynamodb.go
```diff
@@ -440,6 +488,20 @@ func (d DB) DeleteThingWithUnderscores(ctx context.Context, iDApp string) error
...
+}
+
+func toDynamoTimeString(d strfmtTime) string {
+	switch v := d.(type) {
+	case strfmt.DateTime:
+		return time.Time(v).Format(time.RFC3339) // dynamodb attributevalue only supports RFC3339 resolution
+	case *strfmt.DateTime:
+		return time.Time(*v).Format(time.RFC3339) // dynamodb attributevalue only supports RFC3339 resolution
+	default:
+		log.Fatal("oops")
```
natebrennand: I don't think we want to have a fatal log, maybe a `panic`?

samples/gen-go-db/server/db/dynamodb/teachersharingrule.go
```diff
@@ -195,6 +195,68 @@ func (t TeacherSharingRuleTable) getTeacherSharingRulesByTeacherAndSchoolApp(ctx
...
+	} else {
+		queryInput.KeyConditionExpression = aws.String("#TEACHER = :teacher AND #SCHOOL_APP >= :schoolApp")
+	}
+
+	queryOutput, err := t.DynamoDBAPI.QueryWithContext(ctx, queryInput)
+	if err != nil {
+		return err
+	}
+	if len(queryOutput.Items) == 0 {
+		fn(nil, false)
```
rgarcia: curious what's the reasoning here? I'd be surprised receiving a `nil` model in the callback.

hardcoded/hardcoded.go
```diff
@@ -214,8 +214,8 @@ var _bindata = map[string]*asset{
 			"\xf7\x6f\xab\xef\x11\xed\xdd\x8f\x87\x7c\x78\xd5\x41\xfa\x14\x94\x86\x22\x74\x8a\x2c\xa8\xbf\xd8" +
 			"\xde\xa2\x83\xa7\x69\x7c\x74\x96\xfc\x3f\x01\x00\x00\xff\xff",
 		size: 10915,
-		mode: 0664,
-		time: time.Unix(1544125033, 116862280),
+		mode: 0644,
+		time: time.Unix(1553712379, 443321183),
```
natebrennand: there's a flag for `genbin` (I think `--nometadata`) that removes these timestamps

samples/gen-go-db/server/db/interface.go
```diff
@@ -145,6 +163,17 @@ type SchoolApp struct {
 	App    string
 }
 
+// GetTeacherSharingRulesByTeacherAndSchoolAppPageInput is the query input to GetTeacherSharingRulesByTeacherAndSchoolAppPage.
+type GetTeacherSharingRulesByTeacherAndSchoolAppPageInput struct {
+	// StartingAfter is a required specification of an (exclusive) starting point.
+	StartingAfter *models.TeacherSharingRule
```
rgarcia: 1) why is this required? I'd expect an option similar to `Scan` where I don't specify a starting point
2) I think this should be `StartingAt`, not `StartingAfter`, since it is an *inclusive* starting point. I.e. the comparison is >= or <= in the query (depending on descending or not).

server/gendb/interface.go.tmpl
```diff
@@ -35,6 +35,8 @@ type Interface interface {
     {{- $indexName := indexName $xdbConfig.DynamoDB.KeySchema }}
     // Get{{ $modelName }}sBy{{ $indexName }} retrieves a list of {{ $modelName }}s from the database.
     Get{{ $modelName }}sBy{{ $indexName }}(ctx context.Context, input Get{{ $modelName }}sBy{{ $indexName }}Input) ([]models.{{ $modelName }}, error)
+    // Get{{ $modelName }}sBy{{ $indexName }}Page retrieves a page of {{ $modelName }}s from the database.
+    Get{{ $modelName }}sBy{{ $indexName }}Page(ctx context.Context, input Get{{ $modelName }}sBy{{ $indexName }}PageInput, fn func(m *models.{{ $modelName }}, last{{ $modelName }} bool) bool) error
```
natebrennand: asked in Slack too, but I'm unsure about this API of passing a lambda not returning the slice

server/gendb/.notes
```diff
@@ -0,0 +1,3 @@
+thingwithcompositeattributes.go:230  -  is string pointer, needs to be &
+  	S: aws.String(fmt.Sprintf("%s@%s", input.StartingAfter.Name, input.StartingAfter.Branch)),
+
```
rgarcia: you can delete this file :)

samples/gen-go-db/server/db/dynamodb/teachersharingrule.go
aastein: I do not remember. It doesn't make sense.

samples/gen-go-db/server/db/interface.go
aastein: 1. if a string ExpressionAttributeValues is empty Dynamo returns this error 
`ValidationException: One or more parameter values were invalid: An AttributeValue may not contain an empty string`
2. The generated method uses ExclusiveStartKey to avoid the duplicated ending and first result when paging. The comparisons are currently incorrect and were copied from existing methods and are effectively > and <, not >= and <=.

rgarcia: re: 1) it looks like you can omit ExpressionAttributeNames and ExpressionAttributeValues: https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_Query.html (they are both marked required: No)

re: 2) I see... I anticipate that the use case for the existing query methods will be missed, i.e. there are use cases where we want to list out objects from an index starting *at* a specific place. One way you could potentially do this is to accept both `StartingAt` and `StartingAfter`, where both are optional pointers. If `StartingAt` is specified, you populate `ExpressionAttributeNames`, `ExpressionAttributeValues`, `KeyConditionExpression`. If `StartingAfter` is specified, you populate `ExclusiveStartKey`

aastein: nice. I'll try these things.

aastein: I tried these things. It looks like these params are actually required. :fakenews:

No ExpressionAttributeNames and ExpressionAttributeValues
`ValidationException: Invalid KeyConditionExpression: An expression attribute name used in the document path is not defined; attribute name: #DISTRICT
	status code: 400, request id: 90e12efd-f920-4f9b-a429-ea9c5d176c63`

No ExpressionAttributeNames, ExpressionAttributeValues, or KeyConditionExpression
`ValidationException: Either the KeyConditions or KeyConditionExpression parameter must be specified in the request.
	status code: 400, request id: a1a55708-6ab6-4e5a-b195-e64a94d8ede8`

But I can still implement the StartingAt by optionally having the ExclusiveStartKey, everything else seems to be required in practice.

rgarcia: You could enforce that either `StartingAt` or `StartingAfter` is specified (not both), and in the case where `StartingAfter` is specified, use those values in the `KeyConditionExpression` as if they were `StartingAt` (i.e. with the same `>=`/`<=`). It'll be irrelevant since the exclusive start key will be more constraining, but :shrug:

aastein: I went with. That cool?

```
input: {
  StartingAt: model
  Exclusive: bool
}
```

rgarcia:
:thumbsup:

samples/gen-go-db/server/db/dynamodb/thingwithdatetimecomposite.go
```diff
@@ -160,6 +160,69 @@ func (t ThingWithDateTimeCompositeTable) getThingWithDateTimeCompositesByTypeIDA
 	return decodeThingWithDateTimeComposites(queryOutput.Items)
 }
 
+func (t ThingWithDateTimeCompositeTable) getThingWithDateTimeCompositesByTypeIDAndCreatedResourcePage(ctx context.Context, input db.GetThingWithDateTimeCompositesByTypeIDAndCreatedResourcePageInput, fn func(m *models.ThingWithDateTimeComposite, lastThingWithDateTimeComposite bool) bool) error {
+	queryInput := &dynamodb.QueryInput{
```
rgarcia: might be worthwhile to add a check that `StartingAt` and `Limit` (required fields) are non-nil

aastein:
@rgarcia I wouldn't be opposed to either. I think if we left in the deprecated methods but renamed them they wouldn't be used and it would still break existing code. It makes more sense to me to remove them. Should a solution for this also be part of this PR?

rgarcia:
@astein ok then I say go ahead and remove them while we're looking at this code, and then we can publish a breaking change. I can do some follow up with users of x-db so it's not a surprise

aastein:
Other than updating VERSION to v2.0.0 anything this PR needs before merging?

rgarcia:
@aastein nope :shipit:!

Once merged it should trigger the creation of a 2.0 release on github. Once that happens I'd send a note about it to #eng-releases

PR #202, opened 2019-04-16
Fix secondary index code generation bug

aastein:
Code generation would fail for tables which have a primary index without a range key and secondary indexes with range keys.

The code failed when trying to access the non existent range key.
https://github.com/Clever/wag/pull/201/files#diff-7a247ec50143b312e0f88b5e79955fe7R427

This PR moves that code into an if-end block that checks that the primary index has a range key.


aastein:
I'm working on the test now. I'll put it in another PR to get this fix out asap.

aastein:
...and I forgot the version

PR #205, opened 2019-04-17
WIP fix for required hash + range

rgarcia:


rgarcia:
yeah, I also think ideally there's a non-nil check on these values before dereferencing

PR #209, opened 2019-04-18
DynamoDB: convert resource not found to table not found

Sayan-:
https://clever.atlassian.net/browse/INFRA-3340

Did a little bit more digging. Here's the breakdown by method:

Calls that don't need to be updated
* CreateTableWithContext
* ScanPagesWithContext

Calls that need to be updated
* PutItemWithContext
* GetItemWithContext
* QueryWithContext
* DeleteItemWithContext


rgarcia:
Nice!! Lgtm

PR #210, opened 2019-04-19
Support optional range key for dynamodb query input

aastein:
* Support optional range key for dynamodb query input
* Added test based on catapult deployment https://github.com/Clever/catapult/blob/master/swagger.yml#L822

Single attribute range key
```
// GetDeploymentsByEnvAppAndDateInput is the query input to GetDeploymentsByEnvAppAndDate.
type GetDeploymentsByEnvAppAndDateInput struct {
	Environment    string
	Application    string
	DateStartingAt *strfmt.DateTime
	StartingAfter  *models.Deployment
	Descending     bool
	// Limit is an optional limit on how many items to evaluate.
	Limit *int64
}
```

Composite range key
```
// GetTeacherSharingRulesByTeacherAndSchoolAppInput is the query input to GetTeacherSharingRulesByTeacherAndSchoolApp.
type GetTeacherSharingRulesByTeacherAndSchoolAppInput struct {
	Teacher    string
	StartingAt *SchoolApp
	// StartingAfter is a required specification of an exclusive starting point.
	StartingAfter *models.TeacherSharingRule
	Descending    bool
	// DisableConsistentRead turns off the default behavior of running a consistent read.
	DisableConsistentRead bool
	// Limit is an optional limit on how many items to evaluate.
	Limit *int64
}
```


aastein:
Not sure what's going on with the test db. Maybe we're somehow hitting a max connection to the db?

samples/gen-go-db/server/db/dynamodb/dynamodb_test.go
```diff
@@ -22,20 +21,24 @@ func TestDynamoDBStore(t *testing.T) {
...
 	if err := cmd.Start(); err != nil {
 		t.Fatal(err)
 	}
 
-	// wait for dynamodb local to output the correct startup log
-	scanner := bufio.NewScanner(ddbLocalOutputReader)
-	for scanner.Scan() {
-		if strings.Contains(scanner.Text(), "Initializing DynamoDB Local with the following configuration") {
+	end := time.Now().Add(60 * time.Second)
+	var c net.Conn
```
rgarcia: maybe move this into the for loop and close it in the for loop? I'm actually not entirely sure what happens in the case where you overwrite a connection like this...

samples/gen-go-db/server/db/dynamodb/deployment.go
```diff
@@ -0,0 +1,568 @@
...
+		return nil, err
+	}
+
+	return &m, nil
+}
+
+func (t DeploymentTable) getDeploymentsByEnvAppAndVersion(ctx context.Context, input db.GetDeploymentsByEnvAppAndVersionInput, fn func(m *models.Deployment, lastDeployment bool) bool) error {
+	if input.VersionStartingAt != nil && input.StartingAfter != nil {
+		return fmt.Errorf("Must specify one of input.VersionStartingAt or input.StartingAfter")
+	}
```
rgarcia: You should be able to specify neither, i.e. just query on the value of the hash key

samples/gen-go-db/server/db/dynamodb/deployment.go
```diff
@@ -0,0 +1,568 @@
...
+			},
+			"envApp": &dynamodb.AttributeValue{
+				S: aws.String(fmt.Sprintf("%s--%s", input.StartingAfter.Environment, input.StartingAfter.Application)),
+			},
+		}
+		if input.Descending {
+			queryInput.KeyConditionExpression = aws.String("#ENVAPP = :envApp AND #VERSION <= :version")
+		} else {
+			queryInput.KeyConditionExpression = aws.String("#ENVAPP = :envApp AND #VERSION >= :version")
+		}
```
rgarcia: just checking--are these `KeyConditionExpression`s necessary when specifying `ExclusiveStartKey`?

If they are necessary, I think we can simply do `queryInput.KeyConditionExpression = aws.String("#ENVAPP = :envApp")`, since the `<=` or `>=` constraint is less binding than the `ExclusiveStartKey`

samples/gen-go-db/server/db/dynamodb/deployment.go
```diff
@@ -0,0 +1,568 @@
...
+	}
+
+	queryOutput, err := t.DynamoDBAPI.QueryWithContext(ctx, queryInput)
```
rgarcia: I think this needs to be [`QueryPagesWithContext`](https://godoc.org/github.com/aws/aws-sdk-go/service/dynamodb#DynamoDB.QueryPagesWithContext) otherwise for large values of `Limit` we will only get the first page of dynamo results

samples/gen-go-db/server/db/dynamodb/deployment.go
aastein: that's supported. This errors if both are specified.

rgarcia:
oops I misread :facepalm:

samples/gen-go-db/server/db/dynamodb/deployment.go
aastein: ahhh the error message is incorrect tho

samples/gen-go-db/server/db/dynamodb/deployment.go
aastein: nice catch!

samples/gen-go-db/server/db/dynamodb/deployment.go
aastein: very nice catch!

aastein:
@rgarcia I fixed stuff per your suggestions. For the tests failing I "fixed" it by restarting the db if it exits before the tests are done. It workers 🤷‍♂️

aastein:
actually that fix didnt work because the in memory db doesnt retain data across starts, obvi

samples/gen-go-db/server/db/dynamodb/dynamodb-local.sh
```diff
@@ -18,4 +18,4 @@ then
     tar -zxvf /tmp/dynamodb_local_latest.tar.gz -C /tmp/
 fi
 
-exec java -jar "$jar" -sharedDb -inMemory -port 8002
+exec java -jar "$jar" -sharedDb -port 8002
```
rgarcia: what's the reason for this change? I don't remember why i chose `inMemory` originally

samples/gen-go-db/server/db/dynamodb/dynamodb_test.go
```diff
@@ -21,20 +20,41 @@ func TestDynamoDBStore(t *testing.T) {
...
+	testsDone := make(chan struct{})
+	defer close(testsDone)
+	go func(doneC chan struct{}) {
+		cmd := exec.CommandContext(testCtx, "./dynamodb-local.sh")
+		if err := cmd.Start(); err != nil {
+			t.Fatal(err)
+		}
+		for {
+			select {
+			case <-doneC:
```
rgarcia: I think this channel is equivalent to the `testCtx.Done()` channel

samples/gen-go-db/server/db/dynamodb/dynamodb_test.go
```diff
@@ -21,20 +20,41 @@ func TestDynamoDBStore(t *testing.T) {
...
+				return
+			default:
+				if err := cmd.Wait(); err != nil {
```
rgarcia: maybe log the error here for some visibility into how the hack is behaving

samples/gen-go-db/server/db/dynamodb/dynamodb_test.go
```diff
@@ -21,20 +20,41 @@ func TestDynamoDBStore(t *testing.T) {
...
+					cmd = exec.CommandContext(testCtx, "./dynamodb-local.sh")
+					cmd.Start()
```
rgarcia: handle error and `t.Fatal(err)`

samples/gen-go-db/server/db/dynamodb/dynamodb_test.go
```diff
@@ -21,20 +20,41 @@ func TestDynamoDBStore(t *testing.T) {
...
+		c, err = net.Dial("tcp", "localhost:8002")
+		if err == nil {
+			c.Close()
 			break
 		}
+		time.Sleep(time.Second)
+	}
+	if err != nil {
+		t.Fatal(err)
 	}
```
rgarcia: maybe clearer, with one less variable outside the for scope:

```
	// loop for 60s trying to establish a connection
	connected := false
	for start := time.Now(); start.Before(start.Add(60 * time.Second)); time.Sleep(1 * time.Second) {
		if c, err := net.Dial("tcp", "localhost:8002"); err == nil {
			c.Close()
			connected = true
			break
		}
	}
	if connected == false {
		t.Fatal("failed to connect within 60 seconds")
	}
```

samples/gen-go-db/server/db/dynamodb/deployment.go
```diff
@@ -0,0 +1,557 @@
...
+	var decodeErr error
+	var items []models.Deployment
+	pageFn := func(queryOutput *dynamodb.QueryOutput, lastPage bool) bool {
+		if len(queryOutput.Items) == 0 {
+			return false
+		}
+		items, decodeErr = decodeDeployments(queryOutput.Items)
+		if decodeErr != nil {
+			return false
+		}
```
rgarcia: I get a little wierded out by reusing the items array here and putting it in the outer scope, maybe consider

```
		items, err := decodeDeployments(queryOutput.Items)
		if err != nil {
			pageFnErr = err // pageFnErr might be a better name?
			return false
		}
```

samples/gen-go-db/server/db/dynamodb/dynamodb-local.sh
aastein: if the db crashed after tables are created but before queries are sent to the table, the db will be restarted without the tables because they were stored in memory. Removing this option writes the db to a file so it persists across crashes.

PR #212, opened 2019-04-23
DynamoDB range queries: Validate required fields 

aastein:
People are only specifying StartingAt or StartingAfter and not the hash key. This PR add a check that string hash keys are not empty and comments the hash keys so it is more clear that they are required fields.


aastein:
We can do the empty check for string keys now. For number keys the value will never be empty. I think a future API could declare the hash key input fields as pointers to handle all hash key data types.

PR #213, opened 2019-04-30
Address JS client code callback bugs

arsalansufi:
> Commit by commit reviewing recommended

# Overview

This PR addresses two JS client code bugs, both involving callback syntax.

## Fix 1

The client code will now apply the provided callback (if any) to circuit-breaker errors.

Consider the following code sample:

```ts
wagClient.someCall(input, (err, data) => {
  if (err) {
    res.sendStatus(500);
    return;
  }
  res.json(data);
});
```

With the existing client code, if the circuit breaker kicks in, the callback will never actually be called and the request will hang. Now, we'll call the callback and `err` will be populated with the circuit-breaker error.

## Fix 2

The client code will no longer return a rejected promise if the provided callback handles it.

Consider the same code sample as above:

```ts
wagClient.someCall(input, (err, data) => {
  if (err) {
    res.sendStatus(500);
    return;
  }
  res.json(data);
});
```

With the existing client code, even though the callback is handling the error, an unhandled promise rejection will still be generated. That'll no longer be the case.

# Testing

- [x] Made sure that existing JS client code tests still pass
- [x] Added an additional test

# Rollout

This change won't actually make it to production until:

1. A WAG client is regenereated, and
2. A repo pulls that updated WAG client in

### DIFF
--- a/clients/js/genjs.go
+++ b/clients/js/genjs.go
@@ -178,6 +178,21 @@ function responseLog(logger, req, res, err) {
 }
 
 /**
+ * Takes a promise and uses the provided callback (if any) to handle promise
+ * resolutions and rejections
+ */
+function applyCallback(promise, cb) {
+  if (!cb) {
+    return promise;
+  }
+  return promise.then((result) => {
+    cb(null, result);
+  }).catch((err) => {
+    cb(err);
+  });
+}
+
+/**
  * Default circuit breaker options.
  * @alias module:{{.ServiceName}}.DefaultCircuitOptions
  */
@@ -359,29 +374,14 @@ var packageJSONTmplStr = `{
 var methodTmplStr = `
   {{.MethodDefinition}}
     {{if .IterMethod -}}
-    const it = (f, saveResults, cb) => new Promise((resolve, reject) => {
+    const it = (f, saveResults) => new Promise((resolve, reject) => {
     {{- else -}}
     if (!cb && typeof options === "function") {
-      cb = options;
       options = undefined;
     }
 
     return new Promise((resolve, reject) => {
     {{- end}}
-      const rejecter = (err) => {
-        reject(err);
-        if (cb) {
-          cb(err);
-        }
-      };
-      const resolver = (data) => {
-        resolve(data);
-        if (cb) {
-          cb(null, data);
-        }
-      };
-
-
       if (!options) {
         options = {};
       }
@@ -393,7 +393,7 @@ var methodTmplStr = `
       const headers = {};
       {{- range $param := .PathParams}}
       if (!params.{{$param.JSName}}) {
-        rejecter(new Error("{{$param.JSName}} must be non-empty because it's a path parameter"));
+        reject(new Error("{{$param.JSName}} must be non-empty because it's a path parameter"));
         return;
       }
       {{- end -}}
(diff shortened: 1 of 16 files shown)


clients/js/genjs.go
```diff
@@ -569,8 +569,13 @@ var singleParamMethodDefinitionTemplateString = `/**{{if .Description}}
   {{.MethodName}}({{range $param := .Params}}{{$param.JSName}}, {{end}}options) {
   {{- else}}
   {{.MethodName}}({{range $param := .Params}}{{$param.JSName}}, {{end}}options, cb) {
-    return this._hystrixCommand.execute(this._{{.MethodName}}, arguments);
+    let callback = cb;
+    if (!cb && typeof options === "function") {
+      callback = options;
+    }
+    return applyCallback(this._hystrixCommand.execute(this._{{.MethodName}}, arguments), callback);
```
arsalansufi: The core of the fix for both issues listed in the PR description involves lifting the callback application up to the top most level

clients/js/genjs.go
```diff
@@ -177,6 +177,21 @@ function responseLog(logger, req, res, err) {
   }
 }
 
+/**
+ * Takes a promise and uses the provided callback (if any) to handle promise
+ * resolutions and rejections
+ */
```
cozmo: Can you add the jsDoc `@private` tag so as not to expose these functions in the generated README? http://usejsdoc.org/tags-private.html

test/js/circuit_breaker.ts
```diff
@@ -176,5 +176,36 @@ describe("circuit", function() {
...
+        errorD: () => {},
+        infoD: () => {},
+      },
+      retryPolicy: RetryPolicies.None,
+    });
+
+    for (let i = 0; i < 10; i++) {
+      const mockApi = nock(mockAddress).get("/v1/authors").replyWithError("Yikes");
+      // The `await` here is necessary to surface assertion errors in the callback. Without the
+      // `await`, assertion errors will generate unhandled promise rejections and be swallowed
```
cozmo: Instead of mixing callbacks and promises here in a way that is a little confusing I think even with the comment, why not change line 180 to have a `done` callback and call it in the callback once we've asserted.

arsalansufi: Good point! I tried using `done` earlier but ran into some snags. Lemme give it another shot though with a fresh mind

test/js/circuit_breaker.ts
```diff
@@ -192,19 +193,25 @@ describe("circuit", function() {
...
           assert.equal(
-            false, mockApi.isDone(), "Expected circuit breaker to kick in after 5 API calls");
-          assert.equal("OpenCircuitError", err.message);
+            "CommandRejected", err.message,
+            "Expected circuit breaker to short circuit last 5 client calls");
+          assert.equal(false, mockApi.isDone(), "Expected no API call for last 5 client calls");
+        }
+      }).then(() => {
+        // Mark the test as a success if we've made it to the last client call
+        if (i === 9) {
```
arsalansufi: Realized that you can't expect promises to resolve in the order of their creation. Turns out that promises `i === 5 through 9` were resolving before promises `i === 0 through 4` because circuit-breaked promises resolve faster than ones that actually make API calls.

Switched to `Promise.all` accordingly in https://github.com/Clever/wag/pull/213/commits/6f3931faa51c7dc7903d5ff5a4c326d022f9879d

PR #217, opened 2019-05-17
Typescriptify

jefff:
Most important file to review:  `clients/js/genjs.go` 
The `samples` files are auto-generated and may be hepful to look at.


rgarcia:
@jefff can you add a section in the README describing how to use these?

PR #247, opened 2019-07-26
add more useful tracing data

natebrennand:
- configure service name based on `_APP_NAME`
- add `pod_{id,account,region}` tags
- add the request URL's query
- add the requst & response sizes as logs
- add a span for request & response JSON [un]marshaling

Todo:

- [x] Run `make build`
- [x] Run `make generate`


server/genserver.go
```diff
@@ -102,25 +103,25 @@ func (s *Server) Serve() error {
...
-			Type:          jaeger.SamplerTypeRateLimiting,
-			Param:         5,
-			MaxOperations: 100,
-		}
-		cfgTags := []opentracing.Tag{
-			opentracing.Tag{Key: "app_name", Value: os.Getenv("_APP_NAME")},
-			opentracing.Tag{Key: "build_id", Value: os.Getenv("_BUILD_ID")},
-			opentracing.Tag{Key: "deploy_env", Value: os.Getenv("_DEPLOY_ENV")},
-			opentracing.Tag{Key: "team_owner", Value: os.Getenv("_TEAM_OWNER")},
+		samplingRate := .05 // 5%% of requests
```
Sayan-: Is there anyway we could add an upper bound here? Since we're in the early stages I'm concerned our SignalFX gateway won't be able to handle this increased throughput.

natebrennand: Pending feedback from @mohit I'm just going to revert this commit since I have no context on the state of the gateway and the docs around this sampler do not make me confident that I have configured it correctly 🤷‍♂

natebrennand: moved to https://github.com/Clever/wag/pull/249

PR #254, opened 2019-08-16
End paginated queries when Limit is reached

aastein:
Todo: Should we do this for Scan as well?

- [x] Run `make build`
- [x] Run `make generate`


rgarcia:
sorry, just saw your question, but probably makes sense to do it for scan as well

aastein:
Just checked scan. Although AWS supports a limit it doesn't look like we use it so gonna merge this. Maybe later we can add a scan limit.

PR #255, opened 2019-08-21
Separate out NewWithMiddleware

venalen:
This PR splits up `NewWithMiddleware` into `NewRouter` (which will return a `mux.Router`) and `AttachMiddleware` (which attaches the middleware to the http.Handler)

The purpose of this is to be able to add handrolled `gorilla.mux` endpoints which is useful since swagger abstracts the `ResponseWriter` and `Request` away from us. `New` and `NewWithMiddleware` will function the same way.

Todo:

- [x] Run `make build`
- [x] Run `make generate`


server/genserver.go
```diff
@@ -205,15 +205,15 @@ func New(c Controller, addr string) *Server {
 	return NewWithMiddleware(c, addr, []func(http.Handler) http.Handler{})
 }
 
-// NewWithMiddleware returns a Server that implemenets the Controller interface. It runs the
-// middleware after the built-in middleware (e.g. logging), but before the controller methods.
-// The middleware is executed in the order specified. The server will start when "Serve" is called.
-func NewWithMiddleware(c Controller, addr string, m []func(http.Handler) http.Handler) *Server {
+// NewRouter returns a Handler (more specifically mux.Router) with no middleware
+func NewRouter(c Controller) http.Handler {
```
rgarcia: I would change this to return the `*mux.Router` to make it clear that we're exposing it to attach additional routes.

I'd also mention that in the comment as well

server/genserver.go
```diff
@@ -222,11 +222,25 @@ func NewWithMiddleware(c Controller, addr string, m []func(http.Handler) http.Ha
...
+// The middleware is executed in the order specified. The server will start when "Serve" is called.
+func NewWithMiddleware(c Controller, addr string, m []func(http.Handler) http.Handler) *Server {
+	router := newRouter(c)
+
+	return AttachMiddleware(router, addr, m)
 }
-`
+
+// AttachMiddleware attaches the given middleware to the router 
+func AttachMiddleware(router http.Handler, addr string, m []func(http.Handler) http.Handler) *Server {
```
rgarcia: here I'd also change this from `http.Handler` to `*mux.Router` and explain in the comment that it is meant to be used in conjuction with `NewServer`. I might also add in the comment that it attaches custom middleware passed as arguments as well as built-in wag middleware for logging, handling panics, and tracing.

PR #260, opened 2019-10-08
adding pod shortname to tracing tags

ghirsch1:
https://clever.atlassian.net/browse/INFRA-3680

Todo:

- [x] Run `make build`
- [x] Run `make generate`


VERSION
```diff
@@ -1,4 +1,5 @@
-v3.8.0
+v3.8.1
```
Sayan-: nit - semver should be a minor bump

PR #264, opened 2019-10-20
Upgrade go to v1.13

taylor-sutton:


mcab:
Taking a look right now.

I'm assuming that it's not passing checks due to relying on binaries that have been updated for go 1.13, so will try `dep ensure -update` and try again.

mcab:
Upgrading dependencies caused:

```
go generate ./hardcoded/
go generate ./server/gendb/
go build -o bin/wag
# github.com/Clever/wag/vendor/github.com/go-swagger/go-swagger/generator
vendor/github.com/go-swagger/go-swagger/generator/operation.go:116:4: cannot use analyzed.SecurityRequirementsFor(operation) (type [][]analysis.SecurityRequirement) as type []analysis.SecurityRequirement in field value
vendor/github.com/go-swagger/go-swagger/generator/support.go:636:17: cannot use a.Analyzed.SecurityRequirementsFor(o) (type [][]analysis.SecurityRequirement) as type []analysis.SecurityRequirement in assignment
Makefile:19: recipe for target 'build' failed
make: *** [build] Error 2
Exited with code 2
```

Might retry with go 1.13 installed. 😅

mcab:
Hm. Updating binaries didn't work out (from above), but my guess it's coming from needing to update `go-openapi/validate`, which (if we update all of those dependencies), means we have to update `go-swagger` to match new signatures.

Might have to think about this one.

mcab:
It was probably #265 :D

mcab:
With the dependency issue fixed, `wag` probably needs a version bump, and it'll be ready for go 1.13.

PR #266, opened 2019-10-23
Export typescript types

johnhuangclever:
Move the generate typescript types into a `Models` sub-namespace, which will allow them to be access via `ServiceName.Models.Type`

I ran this binary on app-service here https://github.com/Clever/app-service/pull/341

Todo:

- [x] Run `make build`
- [x] Run `make generate`

### DIFF
--- a/clients/js/genjs.go
+++ b/clients/js/genjs.go
@@ -1047,6 +1047,7 @@ func getErrorTypes(s spec.Swagger) ([]string, error) {
 
 func methodDecl(s spec.Swagger, op *spec.Operation, path, method string) (string, error) {
 	returnType, err := ReturnType(s, op)
+	returnType = JSType(fmt.Sprintf("models.%s", returnType))
 	if err != nil {
 		return "", err
 	}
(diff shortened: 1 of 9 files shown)


clients/js/genjs.go
```diff
@@ -1381,7 +1387,13 @@ declare namespace {{.ServiceName}} {
     {{range .ErrorTypes}}
     {{.}}
     {{end}}
-  }
+	}
+
+	namespace Models {
```
jefff: nit: can you change these tabs to spaces so that we avoid mixing tabs/spaces in the outputed TypeScript? I don't think you need to bother with alignment otherwise since it's more trouble than it's worth.

samples/gen-js-db/index.d.ts
```diff
@@ -49,129 +49,134 @@ interface AddressOptions {
...
+    None: RetryPolicy;
+  }
+
+  const DefaultCircuitOptions: CircuitOptions;
+
+  namespace Errors {
+    
+    class BadRequest {
+  message?: string;
+}
```
rgarcia: I'm seeing compilation errors now on code that passes an argument to the constructor of these error types, e.g.
`new ILTokenServiceErrors.NotFound({ message: "stub error" })` now throws an error `Expected 0 arguments, but got 1.`

PR #267, opened 2019-10-28
default to keepalive true for node clients

rgarcia:
This mirrors how Go clients behave, see https://clever.slack.com/archives/C1LPE2A79/p1572179345026300


rgarcia:
We may want to expand this to match the default Go settings (https://golang.org/pkg/net/http/#RoundTripper) using something like https://www.npmjs.com/package/agentkeepalive

PR #268, opened 2019-10-28
Add linguist-generated to samples directory

Sayan-:
Tagging files with `linguist-generated` makes it so they don't auto-expand during code review. Just makes PRs a little easier

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+samples/ linguist-generated


.gitattributes
```diff
@@ -0,0 +1,2 @@
+samples/ linguist-generated
+
```
rgarcia: nit: extra newline

PR #270, opened 2019-11-06
Update Go clients to use http.DefaultTransport

Sayan-:
We've been historically using a shared, zero valued `http.Transport` for all generated Go clients. Go has a `http.DefaultTransport` with some saner defaults. Let's go ahead and standardize on using the default for go clients.

Additionally, we switch from using a `http.Client` per request to a shared instance on the `WagClient`. `http.Client` is threadsafe. This is a minor optimization that I do not think carries risk (_famous last rights right_ 😅)

### DIFF
--- a/clients/go/gengo.go
+++ b/clients/go/gengo.go
@@ -57,7 +57,7 @@ var _ = bytes.Compare
 type WagClient struct {
 	basePath    string
 	requestDoer doer
-	transport   http.RoundTripper
+	client   	*http.Client
 	timeout     time.Duration
 	// Keep the retry doer around so that we can set the number of retries
 	retryDoer *retryDoer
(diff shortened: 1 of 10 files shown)


clients/go/gengo.go
```diff
@@ -69,6 +69,12 @@ type WagClient struct {
 
 var _ Client = (*WagClient)(nil)
 
+// TODO this is a starting point
+var defaultTransport = &http.Transport{
+    MaxIdleConns:          512,
```
rgarcia: This seems high... I'd expect most services to max out at roughly their max number of concurrent requests

clients/go/gengo.go
```diff
@@ -69,6 +69,12 @@ type WagClient struct {
...
+    IdleConnTimeout:       100 * time.Second, // we need to set this to be less than the idle timeout of our ALBs, which is 120 seconds
+}
+
 // New creates a new client. The base path and http transport are configurable.
```
rgarcia: If the default transport performs well then I say just go with that and get out of the business of tweaking these parameters

clients/go/gengo.go
```diff
@@ -57,7 +57,7 @@ var _ = bytes.Compare
 type WagClient struct {
 	basePath    string
 	requestDoer doer
-	transport   http.RoundTripper
+	client   	*http.Client
```
rgarcia: nice

clients/go/gengo.go
Sayan-: I did `2^ones digit of second` 🙈

clients/go/gengo.go
Sayan-: I'm pretty split brained on it tbh. 

short term -> if `client.SetTransport(http.DefaultTransport)` works out well we can piece meal it into the services we care about most

long term -> before baking it into wag it would be nice to do a fair assessment of organizational requirements vs what the defaults provide. Probably not too big of a deal to me

PR #273, opened 2019-12-02
Do not log process metrics when running locally

jish:
_NOTE: I would recommend reviewing commit by commit._

These process metrics are used to monitor our servers in production,
but they make the development logs pretty difficult to read. This will
allow us to actually read the logs from our development servers while
still preserving process monitoring in production.

Todo:

- [x] Run `make build`
- [x] Run `make generate`


samples/gen-go-blog/server/router.go
```diff
@@ -46,10 +46,13 @@ type Server struct {
...
+	tracingToken := os.Getenv("TRACING_ACCESS_TOKEN")
+	ingestURL := os.Getenv("TRACING_INGEST_URL")
+	isLocal := os.Getenv("_IS_LOCAL") == "true"
 
-	go func() {
-		metrics.Log("blog", 1*time.Minute)
-	}()
+	if !isLocal {
+		startLoggingProcessMetrics()
+	}
```
rgarcia: nit: why the extra function call instead of calling go func()ing `metrics.Log()`?

not a nit but actual correctness issue: I'm pretty sure the call to `metrics.Log` still needs to happen in a goroutine since it blocks: https://github.com/Clever/go-process-metrics/blob/95df732d99126d01eaf230349f3b3a5d9fcab7fc/metrics/metrics.go#L19-L69

jish: Doh, I didn't mean to remove the `go` call 🤦‍♂

PR #277, opened 2019-12-04
Add TypeScript error type constructor declarations

jish:
_NOTE: I recommend reviewing commit by commit._

WAG generates a JavaScript client in `gen-js/index.js`, and alongside the client it generates a TypeScript type declaration file `gen-js/index.d.ts`.

Currently the Error object source code conflicts with the Error object type declarations. The Error objects must take one argument, and that argument must contian a `message` key, otherwise a runtime error is thrown. However, the TypeScript type declaration files do no define a constructor for Error objects so we can not construct these objects properly when using TypeScript.

This pull request defines constructors for Error objects in the TypeScript type declaration files.

See the `$ make generate` commit for an example of how this alters the generated type declarations in the `samples/` directory. https://github.com/Clever/wag/pull/277/commits/dc30f8304ea3f8bc6fba7cc7eb8dd7279e026e79

Todo:

- [x] Run `make build`
- [x] Run `make generate`

### DIFF
--- a/samples/gen-js-blog/index.d.ts
+++ b/samples/gen-js-blog/index.d.ts
@@ -71,13 +71,22 @@ declare namespace Blog {
   const DefaultCircuitOptions: CircuitOptions;
 
   namespace Errors {
+    interface ErrorBody {
+      message: string;
+      [key: string]: any;
+    }
+
     
     class BadRequest {
   message?: string;
+
+  constructor(body: ErrorBody);
 }
     
     class InternalError {
   message?: string;
+
+  constructor(body: ErrorBody);
 }
     
   }
--- a/VERSION
+++ b/VERSION
@@ -1,4 +1,5 @@
-v3.13.2
+v3.13.3
+- v3.13.3 Add constructor declarations to TypeScript error models
 - v3.13.1: adds a Canonical-Resource header to clients
 - v3.12.0: Do not log process metrics when running locally
 - v3.11.0: Compress responses
--- a/clients/js/genjs.go
+++ b/clients/js/genjs.go
@@ -1033,11 +1033,11 @@ func getErrorTypes(s spec.Swagger) ([]string, error) {
 				if schema, ok := s.Definitions[typeName]; !ok {
 					errorTypes = append(errorTypes, fmt.Sprintf("class %s {}", typeName))
 				} else if len(schema.Properties) > 0 {
-					errorType, err := asJSType(&schema, "models.")
+					declaration, err := generateErrorDeclaration(&schema, typeName, "models.")
 					if err != nil {
 						return errorTypes, err
 					}
-					errorTypes = append(errorTypes, fmt.Sprintf("class %s %s", typeName, errorType))
+					errorTypes = append(errorTypes, declaration)
 				}
 			}
 		}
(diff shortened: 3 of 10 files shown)


jish:
@rgarcia @johnhuangclever do you think you'll have time to review this pull request, or should I assign it to someone else?

samples/gen-js-blog/index.d.ts
```diff
@@ -71,13 +71,22 @@ declare namespace Blog {
...
 
   namespace Errors {
+    interface ErrorBody {
+      message: string;
+      [key: string]: any;
+    }
+
     
     class BadRequest {
   message?: string;
```
lubalexuan: Not sure if I understand correctly. If we pass in ErrorBody to the constructor which has a mandatory `message` field, why do we define an extra `message?` in custom Error types?

jish: Today WAG generates type declarations for Error Objects. Some of those objects have many attributes, some of them only have a `message`. But all of them must have at least a `message`.

https://github.com/Clever/wag/blob/74da9c8a9f77da8a8eb24b1f22196e6f63067f8d/samples/gen-js-errors/index.d.ts#L80-L87

That is not changing in this pull request. Error Objects can have a `code` or a `foo` or any number of additional properties:

```ts
class MyError {
  message: string;
  code: number;
  foo: string;
}
```

Currently, regardless of the number of attributes an Error Object has, it must be constructed with an object that has a `message` key on it:

https://github.com/Clever/wag/blob/74da9c8a9f77da8a8eb24b1f22196e6f63067f8d/samples/gen-js-errors/types.js#L45-L46

That is also not changing in this pull request.

---

What is changing in this pull request is informing TypeScript of the way that these Error Objects must be constructed.

In order to do that, we first define an interface that satisfies this existing requirement. We create an interface that _must_ have a `message` on it, and may also contain any number of additional properties:

```ts
interface ErrorBody {
  message: string;
  [key: string]: any;
}
```

Then we say, all WAG generated Error Objects, must be constructed with an object that satisfies this interface. There must be one argument, and that argument must have at least a `message` defined on it. The argument _may_ also include any number of additional attributes:

```ts
class MyError {
  message: string;
  foo: string;

  constructor(body: ErrorBody);
}
```

Now TypeScript is aware of the constructor restrictions on WAG Error Objects

```ts
new MyError();   // ❌ invalid -- does not take any arguments
new MyError({}); // ❌ invalid -- does contain a `message`
```

But we are allowed to construct any Error Object with one argument that contains a `message`:

```ts
new MyError({ message: "Hello" });             // ✅ valid -- contains a message
new MyError({ message: "Hello", foo: "bar" }); // ✅ also valid
```

lubalexuan: In this case should we remove the `message?: string;` on custom error types, since we are certain that it has a message var passed into its constructor?

lubalexuan: After talking to Jish offline, decided that we should retain the current way of defining `message` on Error classes.
A note for the future: make `message` field required for Custom errors to avoid confusion. Otherwise, (though not likely) this action will be possible: `BadRequest.message = undefined`.  We probably always want an error message defined for all error types.

PR #271, opened 2019-11-14
Compression

rgarcia:
Turn on compression of responses via `CompressHandler` in github.com/gorilla/handlers

Clients work w/o modification 🙌 

- [ ] Run `make build`
- [ ] Run `make generate`
- [ ] Update the current version in the `/VERSION` file.

PR #278, opened 2019-12-05
Accept gzip

rgarcia:
wag servers have the ability to respond with gzipped results: https://github.com/Clever/wag/pull/271

but clients don't ask for it 🤦‍♂ now they do


rgarcia:
actually nvm looks like Go adds this header automatically

PR #279, opened 2019-12-12
send client version as a request header

rgarcia:
https://clever.atlassian.net/browse/INFRA-3829

Sends client version as a header.

Thought about adding a `x-accept-version` swagger configuration + middleware to check it, but stopped short of it for now and settled on a new exported middleware that can be opted into / configured when constructing wag servers. If a lot of wag servers end up using it I think it might make sense to add the swagger extension.

In the process of coming up with the javascript client changes my editor removed whitespace on empty lines, so I went ahead and fixed that too.


jish:
Nice addition 🙂 I'm curious why we didn't use `User-Agent` for this? 🤔 

    User-Agent: Node/12.0 AppServiceClient/1.2.3

Does it feel like that would be a mis-use of `User-Agent`?

rgarcia:
yeah I didn't really want to mess with parsing user agent strings

PR #281, opened 2019-12-20
Allow multiple success cases

jish:
We want to create an endpoint that will idempotently add a new
resource. If the request is sent twice in a row, that is ok. It will
create the resource on the first attempt, and respond with `201 Created`.
On the second attempt, it will respond with `200 OK`.

Todo:

- [x] Run `make build`
- [x] Run `make generate`

### DIFF
--- a/VERSION
+++ b/VERSION
@@ -1,4 +1,5 @@
-v3.14.1
+v3.15.0
+- v3.15.0 Allow multiple success responses
 - v3.14.1 Send client version as a header bugfix.
 - v3.14.0 Send client version as a header.
 - v3.13.3 Add constructor declarations to TypeScript error models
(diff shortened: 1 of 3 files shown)


cozmo:
Wait please don’t merge this without a shepherd. This breaks a core model of wag and I see no tests or even proof of concept that it works. 

How do you use this? As in how does the codegen work to allow a function to return 2 different types, and how do clients tell the difference?

cozmo:
Ah I see now that Raf is also assigned, which mobile GitHub didn’t make obvious. Concern still stands but we have the right eyes 👍

rgarcia:
@jish what's your understanding of why this restriction exists in the first place?

Without digging through the code and dusting off my memory, I think this breaks the assumption of go client method signatures having one type to return on success, e.g.

`successResponse, err := client.Operation(ctx, input)`

The codegen doesn't have a way of dealing with `successResponse` taking on different forms.

PR #283, opened 2019-12-30
clean up unused code

prime-time:
Todo:

- [x] Run `make build`
- [x] Run `make generate`

While I was looking through the codebase to get a better feel for wag, I noticed some unused functions in `main.go` that had been left behind from a refactor https://github.com/Clever/wag/commit/05821f664aa4d898d14dcf5d940003dc75373c14#diff-f8456eec252286d7dd7c40862aef0c49.

I also see a bunch of glide validation code, but I wonder if we could also remove that since we don't use glide anymore https://github.com/Clever/wag/blob/cf9f5de21695f155fd2d2c57a99a8846fd4d52ad/main.go#L49-L65

### DIFF
--- a/VERSION
+++ b/VERSION
@@ -1,4 +1,5 @@
-v3.14.2
+v3.14.3
+- v3.14.3 Remove unused code.
 - v3.14.2 Stricter span checks to prevent tracing errors.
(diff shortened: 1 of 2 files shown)


prime-time:
♻️ 🌳 🌲

PR #286, opened 2020-01-08
Enable compression for js clients

Sayan-:
In https://github.com/Clever/wag/pull/271 we enabled server side gzipping. For golang this is handled directly in the `http.Transport` layer.

Unfortunately the `request` library we use for js doesn't enable gzip by default:

```
For backwards-compatibility, response compression is not supported by default.
To accept gzip-compressed responses, set the gzip option to true.
Note that the body data passed through request is automatically decompressed
while the response object is unmodified and will contain compressed data if
the server sent a compressed response.
```

(from https://www.npmjs.com/package/request#requestoptions-callback)

Todo:

- [x] Run `make build`
- [x] Run `make generate`


clients/js/genjs.go
```diff
@@ -445,7 +445,8 @@ const methodTmplStr = `
 
       const requestOptions = {
         method: "{{.Method}}",
-        uri: this.address + "{{.PathCode}}",
+		uri: this.address + "{{.PathCode}}",
+		gzip: true,
```
rgarcia: nit leading whitespace looks different

Sayan-: Hmm shoot yeah totally missed this - thanks for the catch!

PR #288, opened 2020-01-11
Re-finagle panic recovery

Sayan-:
We have observed the PanicMiddleware can fail to catch all panics. One particular case is a user injects a `http.TimeoutHandler` middleware. Using this middleware causes a request to finish, but the server code
called by the route handler will continue execution and can panic. This panic "post-request" will not be caught by our outermost PanicMiddleware since that handler has already returned

TODO:
- [x] Run `make build`
- [x] Run `make generate`


samples/gen-go-blog/server/router.go
```diff
@@ -157,7 +170,8 @@ func withMiddleware(serviceName string, router http.Handler, m []func(http.Handl
 		handler = m[i](handler)
 	}
 	handler = TracingMiddleware(handler)
-	handler = PanicMiddleware(handler)
+	// logs any panics encountered in our middleware
+	handler = PanicMiddleware(handler, false)
```
aastein: Why should we keep this middleware here is the one above does the same stuff?

Sayan-: This outer middleware provides value by logging panics from any other middlewares. I think it's worth having given the dependency injection pattern for user defined middleware

PR #289, opened 2020-01-30
dynamodb codegen: support binary type

rgarcia:
https://clever.atlassian.net/browse/INFRANG-21

Allows you to create a schema + ddb indexes like
```yaml
  Event:
    type: object
    x-db:
      AllowOverwrites: true
      DynamoDB:
        AttributeDefinitions:
        - AttributeName: "pk"
          AttributeType: "S"
        - AttributeName: "sk"
          AttributeType: "S"
        - AttributeName: "data"
          AttributeType: "B"
        KeySchema:
        - AttributeName: "pk"
          KeyType: HASH
        - AttributeName: "sk"
          KeyType: RANGE
        GlobalSecondaryIndexes:
        - IndexName: bySK
          Projection:
            ProjectionType: ALL
          KeySchema:
          - AttributeName: sk
            KeyType: HASH
          - AttributeName: data
            KeyType: RANGE
        TimeToLiveSpecification:
          AttributeName : ttl
          Enabled: true
    properties:
      pk:
        type: string
      sk:
        type: string
      data:
        type: string
        format: byte
      ttl:
        type: integer
```


rgarcia:
- `type:string ... format:byte` is what go-swagger uses to generate types that contain bytes: https://github.com/go-swagger/go-swagger/blob/master/generator/formats.go

- Yes, since at least with the analogous `format: date-time`, I've found it cumbersome to deal with `strfmt.DateTime` when constructing inputs to the db methods. I think it's more appropriate for the `strftmt` fields to remain as fields on types in the `models` package, since those models get formatted / marshaled. But in the `db` package, it's easier to construct inputs that are the normal types

taylor-sutton:
🚢

PR #290, opened 2020-02-11
Add optional configuration parameters.

bstein-clever:
Inspired by https://dave.cheney.net/2014/10/17/functional-options-for-friendly-apis

Also add support for compression level configuration.


bstein-clever:
This should be backwards compatible without any breakage.
I'm not sure if there's a version bump I need to make.

PR #295, opened 2020-03-09
Generate TypeScript types for allOf

adamvictorclever:
I noticed the generated types for `workflow-manager` had `Workflow` defined as `any` https://github.com/Clever/workflow-manager/blob/master/gen-js/index.d.ts#L261. When I checked I saw that the swagger.yml defines `Workflow` using `allOf` (https://github.com/Clever/workflow-manager/blob/master/swagger.yml#L429-L442), which our TypeScript codegen didn't support. If my understanding of `allOf` is correct, it basically joins multiple definitions into a single one, which seems pretty useful!

This PR adds codegen of TypeScript types for allOf by creating an intersection type combining all of the subtypes together. See ea00b65 for examples of the generated types

- [X] Run `make build`
- [X] Run `make generate`


adamvictorclever:
Well, it's been over a year since I originally opened this and then let it die 😅 But I recently found myself wanting this again, so I'm re-opening this! I've re-done my changes to make sure they're based on the latest master.

clever-dan-torczynski (approved):
LGTM! Haven't updated wag before so not sure if there are other rollout steps than just updating the version file, might be worth checking with infra.

tnsardesai (approved):
LGTM. Just updating `VERSION` is fine. no extra steps required

PR #303, opened 2020-07-22
JS Client iterator: make forEach work with async functions

aastein:
Trying to do async tasks inside of a wag iter, like so, does not work

```js
const someIter = someService.getSomeIter(params);

const f = async (data) => {
  await someService.doSomething(data);
}

await someIter.forEach(f);
```
because the implementation of `someIter.forEach` does not `await` calls to `f`. 

This PR adds a new method called `forEachAsync` which will wait calls to `f`

the new way uses a `for` loop to preserve part of the behavior of `forEach`: `arr.forEach(callback(currentValue [, index [, array]])[, thisArg])`, which in addition to the current value, the index and entire result set are passed to `f`

- [x] Run `make build`
- [x] Run `make generate`


johnhuangclever:
Heads up, this does not technically match how forEach works on js arrays. It should not be blocking on async functions like for loops:

[Playground Link](https://www.typescriptlang.org/play/?ssl=1&ssc=1&pln=9&pc=19#code/MYewdgzgLgBAtgUwhAhgcwQLhtATgSzDRgF4YByACwQBsaQYB3EXGgE3IG4AoUSEGggB09NAApEydAgCUPXuGgwUpGAG0AjABoYAJh0BmHQBYdAVh0A2HQHYdADgC6PFEIBmLAKIpglMSggATzBgGH8ZUgA+GABvbhgEnAQoABV8RBAAVygxMQiSaL4IAWFRcJ0NAAZquW4AX1qFfkEREHEAIkCQdrkgA)

In this case it should be still possible to use `iter.toArray` and then for loop it there.

aastein:
true, I don't like how js forEach works. I viewed this as an improvement because I wanted to be able to keep a small memory footprint when doing async work for each member of an array. Should we introduce a new method instead? I think having a simple interface to achieve doing async stuff mid-iter is something good to have.

Sayan-:
Ohhh great call out john. `asyncForEach` or `forEachAsync` doesn't seem too crazy to me

aastein:
sgtm! I'll make a new method for this n call it forEachAsync, so when people type `forEa` it will autocomplete / show both options if someone's autocomplete only works on prefixes.

clients/js/genjs.go
```diff
@@ -619,6 +631,7 @@ const pluralParamMethodDefinitionTemplateString = `/**{{if .Description}}
    * @returns {function} iter.map - takes in a function, applies it to each resource, and returns a promise to the result as an array
    * @returns {function} iter.toArray - returns a promise to the resources as an array
    * @returns {function} iter.forEach - takes in a function, applies it to each resource
+   * @returns {function} iter.forEachAsync - takes in a, async function, applies it to each resource
```
Sayan-, line 634: typo

PR #307, opened 2020-11-05
v4.0.0: import package with correct module version

aastein:
v4.0.0: import package with correct module version

new CLI flag:
-output-path

assumptions:
the code assumes it is being run in the same directory as the go.mod file

usage:
```
wag -output-path gen-go -js-path ./gen-js -file swagger.yml 
```
Changes to exported types, methods, and functions:
* does not break existing uses of v3 CLI
* redefines swagger/swagger.go `type Generator struct`
* changes function signature of 
  * clients/go/gengo.go  `func Generate`
  * models/genmodels.go `func Generate`
  * server/gendb/gendb.go `func GenerateDB `

- [x] Run `make build`
- [x] Run `make generate`

### DIFF
--- a/main.go
+++ b/main.go
@@ -50,7 +50,16 @@ func main() {
 		goPackagePath = *goPackageName
 	} else {
 		defer modFile.Close()
-		goPackagePath = getModulePackagePath(*outputPath)
+
+		goPath := os.Getenv("GOPATH")
+		if goPath == "" {
+			log.Fatalf("GOPATH must be set")
+		}
+		if *outputPath == "" {
+			log.Fatal("output-path is required")
+		}
+
+		goPackagePath = getModulePackagePath(goPath, *outputPath)
 		*goPackageName = getModulePackageName(goPackagePath, modFile)
 	}
 
@@ -140,17 +149,10 @@ func main() {
 	}
 }
 
-func getModulePackagePath(outputPath string) string {
+func getModulePackagePath(goPath, outputPath string) string {
 	pwd, err := os.Getwd()
 	if err != nil {
 		log.Fatalf("Error getting current directory: %s", err.Error())
-	}
-	goPath := os.Getenv("GOPATH")
-	if goPath == "" {
-		log.Fatalf("GOPATH must be set")
-	}
-	if outputPath == "" {
-		log.Fatal("output-path is required")
 	}
 	goSrcPath := fmt.Sprintf("%v%v", goPath, "/src/")
 	return path.Join(strings.TrimPrefix(pwd, goSrcPath), outputPath)
--- a/main.go
+++ b/main.go
@@ -158,6 +158,11 @@ func getModulePackagePath(goPath, outputPath string) string {
 	return path.Join(strings.TrimPrefix(pwd, goSrcPath), outputPath)
 }
 
+// getModulePackageName gets the package name of the generated code
+// Example: if packagePath = github.com/Clever/wag/gen-go and the module name is github.com/Clever/wag/v2
+// the function will return github.com/Clever/wag/v2/gen-go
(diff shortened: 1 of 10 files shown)


main.go
```diff
@@ -25,14 +27,34 @@ func main() {
...
+	} else if err == nil {
+		isModulesRepo = true
+	}
+
+	if isModulesRepo {
+		if *goPackagePath == "" {
+			log.Fatal("go-package-path is required")
+		}
+		if *goServicePackageName == "" {
+			log.Fatal("go-package-path is required")
```
ghirsch1, line 55:
```suggestion
			log.Fatal("go-service-package is required")
```

ghirsch1:
Can't we instead just generate everything from `goPackageName` and the major version? Since it'll always be
```
tokens := strings.Split(goPackageName, "/")
goPackagePath := fmt.Sprintf("%v/%v/%v", tokens[:3], semverVersion.Major(), tokens[3:])
```

main.go
```diff
@@ -92,31 +111,72 @@ func main() {
...
+		log.Fatalf("invalid module path %q", modulePath)
+	}
+	pseudoMajor := module.PathMajorPrefix(pathMajor)
+
+	// add module version to package path
+	if pseudoMajor != "" {
+		suffix := strings.TrimPrefix(packagePath, modulePath)
+		return fmt.Sprintf("%v/%v%v", modulePath, pseudoMajor, suffix)
+	}
+	return modulePath
```
ghirsch1, line 181: haven't had a chance to dig into this but am confused by it, probably some comments about what the inputs/outputs are would be helpful

main.go
```diff
@@ -25,16 +29,31 @@ func main() {
...
+	var goPackagePath string
+	// check if the repo uses modules
+	if modFile, err := os.Open("go.mod"); err != nil {
+		if _, ok := err.(*os.PathError); !ok {
+			log.Fatalf("Error checking if the repo contains a go.mod file: %s", err.Error())
+		}
+		if *goPackageName == "" {
+			log.Fatal("go-package is required")
+		}
+		goPackagePath = *goPackageName
```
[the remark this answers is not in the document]
aastein: Only an os.PathError should lead to that path (line 47)

- if the error is not nil and not a os.PathError, then we error out because is is an unknown error
- if the error is not nil and a os.pathError, that means there is no go.mod file and execution continues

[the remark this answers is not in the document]
ghirsch1: oops, I'm stupid and didn't see the `!`

main.go
aastein: let me know if the new function comment helps d4f02d2

main.go
```diff
@@ -92,31 +120,70 @@ func main() {
...
+		log.Fatalf("invalid module path %q", modulePath)
+	}
+	pseudoMajor := module.PathMajorPrefix(pathMajor)
+
+	// add module version to package path
+	if pseudoMajor != "" {
+		suffix := strings.TrimPrefix(packagePath, modulePath)
+		return fmt.Sprintf("%v/%v%v", modulePath, pseudoMajor, suffix)
+	}
+	return modulePath
```
ghirsch1, line 188: Is there a case that this would miss? With `outputPath` being the relative path passed in by the flag
```suggestion
	return fmt.Sprintf("%v/%v", moduleName, outputPath)
```

aastein: lol, jeez. Yeah, that should always work.

ghirsch1 (approved):
okay cool, that makes much more sense now with the comment and update

johnhuangclever:
Hey, do we need to update this as well? https://github.com/Clever/dev-handbook/blob/master/make/wag.mk

PR #311, opened 2020-12-08
Fix scans on DynamoDB indexes with composite attributes

prime-time:
While I was working on making a change to add support for limiting the number of items in DynamoDB index scans, I noticed that the `exclusiveStartKey` wasn't working when scanning an index with a composite attribute. This change aims to fix that, and also makes various other changes to get the tests to pass.

I recommend viewing the changes by commit. I tried to make the commit messages as informative as possible, and I regenerated the generated code after each change so that the changes can be viewed separately.

**The core changes are these:** 
When forming `exclusiveStarKey` as part of the scan input, we need to provide the mapping for composite attributes because the `dynamodbattribute.MarshalMap` function doesn't handle composite attributes.
https://github.com/Clever/wag/blob/a9b710a2f10a0cedf85531b0b409113db8072fc4/samples/gen-go-db/server/db/dynamodb/thing.go#L211

For scans on secondary indexes, both the primary key and secondary index attributes need to be provided in `exclusiveStarKey`, otherwise DynamoDB throws 400 errors.
https://stackoverflow.com/questions/40988397/dynamodb-pagination-with-withexclusivestartkey-on-a-global-secondary-index

**Additional changes to fix the tests:**
I updated the config so that tests are generated for all samples. I then refactored the scan tests because they weren't working on indexes with composite attributes. The tests had been relying on the items returned by the scan in a specific order, however DynamoDB doesn't order the results of scan operations. I also had to make various other changes to get the tests to pass.

- [x] Run `make build`
- [x] Run `make generate`


server/gendb/table.go.tmpl
```diff
@@ -236,15 +236,23 @@ func (t {{ $modelName }}Table) scan{{ $modelName }}s(ctx context.Context, input
     ConsistentRead: aws.Bool(!input.DisableConsistentRead),
   }
   if input.StartingAfter != nil {
+  {{- if indexContainsNonCompositeAttribute $xdbConfig $xdbConfig.DynamoDB.KeySchema }}
     exclusiveStartKey, err := dynamodbattribute.MarshalMap(input.StartingAfter)
     if err != nil {
       return fmt.Errorf("error encoding exclusive start key for scan: %%s", err.Error())
     }
+  {{- end }}
```
prime-time, line 244: Without this condition, `exclusiveStartKey` is unused when the index consists only of composite attributes, and the Go compiler loudly complains about that.

server/gendb/tests.go.tmpl
```diff
@@ -453,34 +453,6 @@ func Get{{ $modelName }}sBy{{ $indexName }}(d db.Interface, t *testing.T) func(t
...
-  err := g.d.Scan{{ $modelName }}s(g.input.ctx, g.input.input, func(m *models.{{ $modelName }}, last bool) bool {
-    {{ varname $modelName }}s = append({{ varname $modelName }}s, *m)
-    return true
-  })
-  var errStr string
-  if err != nil {
-    errStr = err.Error()
-  }
-  require.Equal(t, g.output.err, err, errStr)
-  require.Equal(t, g.output.{{ varname $modelName }}s, {{ varname $modelName }}s)
```
ghirsch1, line 482: Why change the whole format of how these tests are written, rather than just update these assertions?

ghirsch1 (approved):
This generally looks reasonable, aside from the comment about the test refactoring seeming like overkill (and also moving away from the table test format that Clever generally uses for tests in Go).

server/gendb/tests.go.tmpl
prime-time: I was able to get the basic scan test working in the old format by just changing the assertion to not care about the order of the results from the scan. The basic scan tests return all 3 items in the db, so we can use `assert.ElementsMatch` to assert that the result has the same items as what's expected in any order https://godoc.org/github.com/stretchr/testify/assert#ElementsMatch.

However, I wasn't sure how to do that with the tests for starting after. Since we don't know how the items are ordered in the db, if we scan for items starting after x, we don't know how many items there after after x since x can be the first, last, or anything in between. The way I did it was to find the first item, and then scan with starting after the first item so that we know how many items to expect.

Any suggestions on how we could make it work in the existing test format?

ghirsch1: Ah I see, that makes sense. Yeah I don't really have a better idea.

Also since all of these only generate 1 test case each, table tests aren't actually that helpful/different, so there's actually not much benefit to using that structure now that I think about it more carefullly

prime-time: I can add in a comment in the codegen template to provide some context as to why the scan test is structured differently than everything else.

PR #312, opened 2020-12-08
Support limiting the number of items returned by DynamoDB scan operations

prime-time:
This change exposes a new optional `Limit` parameter for generated scan methods which allows scanning up to `Limit` number of items instead of every item. The motivation is that I want to create a new API endpoint which retrieves all items from a DynamoDB table, however there are too many items to do a full table scan in a single request. With this change, I'll be able to implement paging using scan.

The `Limit` parameter introduced here works basically the same way as `Limit` does for queries. I literally copied that code ©️ 😄 .
https://github.com/Clever/wag/blob/a9b710a2f10a0cedf85531b0b409113db8072fc4/server/gendb/interface.go.tmpl#L120-L121
https://github.com/Clever/wag/blob/a9b710a2f10a0cedf85531b0b409113db8072fc4/server/gendb/table.go.tmpl#L434-L438

For reference, this describes how `Limit` works in `ScanPages`:
https://docs.aws.amazon.com/sdk-for-go/api/service/dynamodb/#ScanInput

I manually tested this change on an API endpoint that I'm working on, and it works as expected.

I also added in a comment mentioning why the scan tests are structured differently which came up in #311 

I recommend viewing the changes by commit.

- [x] Run `make build`
- [x] Run `make generate`


ghirsch1:
Is this a one-off/infrequent thing? If so, maybe better to just export the table to S3 and use that? https://aws.amazon.com/blogs/aws/new-export-amazon-dynamodb-table-data-to-data-lake-amazon-s3/

rgarcia (approved):
thanks for contributing this and the bugfix!

rgarcia:
also re:
>  The motivation is that I want to create a new API endpoint which retrieves all items from a DynamoDB table, however there are too many items to do a full table scan in a single request. With this change, I'll be able to implement paging using scan.

generally speaking scans in the request path are not ideal. Querying an index is preferred. You might to share details of the use case in #eng-infra / #dynamodb slack to make sure you won't have a bad time™

prime-time:
The use case for the new endpoint I'm working on is for a script and potentially others in the future, so this wasn't something I was planning on using in application code. I posted the full story in #dynamodb (apologies for the novel 📚 ).

prime-time:
We had a follow up discussion about this via slack, and there were no objections so I'll go forward with merging this 😄

PR #313, opened 2020-12-16
Convert to Modules

aastein:
- Converts the repo to use modules.
- bumps to v5.0.0 so all releases of v5 will be compatible with modules.

Todo:

- [x] Run `make build`
- [x] Run `make generate`


tnsardesai (approved):
LGTM 🚀 🚀 🚀

PR #317, opened 2021-01-25
Client: Add status code to error when status code not recognized

aastein:
When the status code returned by the server is not recognized, the client classifies the error as a `models.InternalError` with the message `"Unknown response"`

This PR changes the message of the error to include the unrecognized status code. For example, `"Unknown status code 503"`

- [x] Run `make build`
- [x] Run `make generate`

### DIFF
--- a/clients/go/gengo.go
+++ b/clients/go/gengo.go
@@ -600,7 +600,7 @@ func parseResponseCode(s *spec.Swagger, op *spec.Operation, capOpID string) stri
 	errorType, _ := swagger.OutputType(s, op, 500)
 	buf.WriteString(fmt.Sprintf(`
 	default:
-		return %s&%s{Message: "Unknown response"}
+		return %s&%s{Message: fmt.Sprintf("Unknown status code %%%%%%%%v", resp.StatusCode)}
 	}
 }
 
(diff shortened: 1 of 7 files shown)


clients/go/gengo.go
```diff
@@ -600,7 +600,7 @@ func parseResponseCode(s *spec.Swagger, op *spec.Operation, capOpID string) stri
 	errorType, _ := swagger.OutputType(s, op, 500)
 	buf.WriteString(fmt.Sprintf(`
 	default:
-		return %s&%s{Message: "Unknown response"}
+		return %s&%s{Message: fmt.Sprintf("Unknown status code %%%%%%%%v", resp.StatusCode)}
```
ghirsch1, line 603: lol what's the conversion happening here that requires `%%%%%%%%v` -> `%v`?

ghirsch1 (approved):
lgtm!

tnsardesai (approved):

clients/go/gengo.go
aastein: this string gets formatted a bunch of times so I needed to escape the `%v` as many times as it's being formatted

aastein: I almost gave up and just used `+`

PR #318, opened 2021-01-28
Fix edge case where composite attribute is used multiple times as a key for an index

dengjulie:
There are duplicate type definitions when a composite attribute is used multiple times as the range key for an index. This PR moves the logic for these struct definitions to be calculated per composite attribute rather than per index. I also ended up finding a bug with the tests - not all of an object's required fields are specified for saving to the db, which causes null pointer dereferences. This specifically happens for a the required fields for composite attribute since we have to access the fields when creating the attribute. 

- [x] Run `make build`
- [x] Run `make generate`


samples/gen-go-db/server/db/interface.go
```diff
@@ -845,6 +856,71 @@ func (e ErrThingWithCompositeAttributesAlreadyExists) Error() string {
...
+	Descending    bool
+	// Limit is an optional limit of how many items to evaluate.
+	Limit *int64
+}
+
+// PropertyOnePropertyTwo struct.
+type PropertyOnePropertyTwo struct {
+	PropertyOne string
+	PropertyTwo string
+}
```
dengjulie, line 908: Problem is that this is redefined

aastein (approved):
LGTM :shipit:

samples/gen-go-db/server/db/interface.go
```diff
@@ -1135,12 +1146,6 @@ type GetThingWithMatchingKeyssByAssocTypeIDAndCreatedBearInput struct {
 	Limit *int64
 }
 
-// CreatedBear struct.
-type CreatedBear struct {
-	Created strfmt.DateTime
-	Bear    string
-}
-
```
prime-time, line 1143: I was concerned for a second when I saw that this bear struct was deleted, but I see that it's just moved down 😌 .

dengjulie: worry not 😌 😌 😌

PR #319, opened 2021-01-29
test example

dengjulie:
Todo:

- [ ] Run `make build`
- [ ] Run `make generate`


samples/gen-go-db/server/db/tests/tests.go
```diff
@@ -5345,6 +5356,331 @@ func ScanThingWithCompositeAttributesRepeatedsByTresAndUnoDos(d db.Interface, t
...
+}
+
+func GetThingWithCompositeAttributesRepeatedsByCuatroAndTres(d db.Interface, t *testing.T) func(t *testing.T) {
+	return func(t *testing.T) {
+		ctx := context.Background()
+		require.Nil(t, d.SaveThingWithCompositeAttributesRepeated(ctx, models.ThingWithCompositeAttributesRepeated{
+			Cuatro: db.String("string1"),
+			Tres:   db.String("string1"),
+			Uno:    db.String("string1"),
+		}))
```
dengjulie, line 5398: The issue was that the created struct did not include all composite keys. When we save the fields into the DB, we build the composite keys and reference the attributes (they must not be null) in order to do so, in this case it is uno and dos, but dos is not defined

dengjulie:
This was a test PR to see whether the bug was already existing or introduced in https://github.com/Clever/wag/pull/318 - it was already existing. Went ahead and fixed the bug in that PR as well.

PR #320, opened 2021-01-29
Remove required attributes from test object initialization

dengjulie:
A better fix to the bug I previously found (and tried to fix) where not all necessary fields are included in initialization and we potentially run into null pointer dereference issues in tests.

Saw the below generated code in another service: seems like there's a lot of incompatibility with adding in lots of new fields if we include all required fields

This issue doesn't actually cause errors since these are only tests called internally, but I thought it would be worth the change since people may eventually add more tests for other cases.

In gen-go/server/db/tests/tests.go of family-invites-service:

image

- [x] Run `make build`
- [x] Run `make generate`


aastein (commented):
in 6.0.1 was is possible to generate invalid tests?

dengjulie:
Yes, here's an example https://github.com/Clever/wag/pull/319/files

dengjulie:
It was causing errors for my PR if I didn't fix it as well

aastein (approved):
LGTM

PR #321, opened 2021-02-01
new flags: --client-only & --client-language

aastein:
This PR adds two command line flags to Wag, `--client-only` and `--client-language`

**`--client-only`** When used wag does not generate server code

**`--client-language`** Can be `go` or `js`. When used wag will generate only the client of the specified language. 
 
**Both flags can be used together or separately.**  When using both flags like so `--client-only --client-language js`, no golang models are generated.

- [x] Run `make build`
- [x] Run `make generate`


main.go
```diff
@@ -22,49 +23,49 @@ import (
...
+	// flag values
+	swaggerFile    *string
+	goPackageName  *string
+	outputPath     *string
+	jsModulePath   *string
+	versionFlag    *bool
+	clientOnly     *bool
+	clientLanguage *string
+
+	// dervied values
```
ghirsch1, line 37:
```suggestion
	// derived values
```

main.go
```diff
@@ -176,3 +217,72 @@ func getModulePackageName(modFile *os.File, outputPath string) string {
 	moduleName = strings.TrimSpace(moduleName)
 	return fmt.Sprintf("%v/%v", moduleName, outputPath)
 }
+
+// validate validates the configuration and determines the goPackagePath for the generated code
+func (c *config) validate() error {
```
ghirsch1, line 222: nit, but could rename to `validateAndSetDerivedFields` to convey that it's not just validating

main.go
```diff
@@ -176,3 +217,72 @@ func getModulePackageName(modFile *os.File, outputPath string) string {
...
+		c.goPackagePath = getModulePackagePath(goPath, *c.outputPath)
+		*c.goPackageName = getModulePackageName(modFile, *c.outputPath)
+	}
+
+	// parse flags determing if client/server code is generated
+	c.generateServer = true
+	c.generateGoClient = true
+	c.generateJSClient = true
+	// only generate client
+	if c.clientOnly != nil && *c.clientOnly == true {
```
ghirsch1, line 253: nit, we often use this style
```suggestion
	if swag.BoolValue(c.clientOnly) {
```

main.go
```diff
@@ -176,3 +217,72 @@ func getModulePackageName(modFile *os.File, outputPath string) string {
...
+		c.generateServer = false
+	}
+
+	// only generate client of specific language
+	if c.clientLanguage != nil && len(*c.clientLanguage) > 0 {
```
ghirsch1, line 258: same nit
```suggestion
	if swag.StringValue(c.clientLanguage) != "" {
```

main.go
```diff
@@ -176,3 +217,72 @@ func getModulePackageName(modFile *os.File, outputPath string) string {
...
+		default:
+			return fmt.Errorf("client-language must be one of \"go\" or \"js\"")
+		}
+	} else {
+		c.generateGoClient = true
+		c.generateJSClient = true
+	}
+
+	c.generateGoModels = c.generateServer || c.generateGoClient
+	if c.generateJSClient && (c.jsModulePath == nil || *c.jsModulePath == "") {
```
ghirsch1, line 278:
```suggestion
	if c.generateJSClient && swag.StringValue(c.jsModulePath) == "" {
```

ghirsch1 (approved):
Looks awesome!

main.go
aastein: just gonna call it `setDerivedFields` because it returns an error so validation is somewhat implied.

PR #327, opened 2021-02-05
Dynamo codegen

aastein:
This PR adds two command line flags to Wag, `-dynamo-only` and` -dynamo-path`

**`-dynamo-only`**  When used wag only generates dynamo code and go models

**`-dynamo-path`** Optional path relative to go package path to generate dynamo code

Both flags can be used together or separately. 

Other:
removed `samples/gen-go-deprecated/README.md`. It is a stale file that is no longer generated but does not get automatically deleted because the dir it is in is persisted when generating code with wag.

Todo:
- [x] Run `make build`
- [x] Run `make generate`
- [todo: is major version bump because exported funcs changed] Update the current version in the `/VERSION` file.


taylor-sutton (approved):
LGTM

aastein:
@taylor-sutton I refactored how the config struct is set. LMK what you think.

main.go
```diff
@@ -215,80 +214,80 @@ func prepareDir(dir string) error {
...
+		if goPackageName == "" {
 			return fmt.Errorf("go-package is required")
 		}
-		c.goPackagePath = *c.goPackageName
+		// if the repo does not use modules, the package name is equivalent to the package path
+		c.goPackagePath = goPackageName
 	} else {
 		defer modFile.Close()
-
+		// todo: do not rely on GOPATH when repo uses modules
```
taylor-sutton, line 269: Extreme-level nit: Should capitalize `TODO`

main.go
```diff
@@ -215,80 +214,80 @@ func prepareDir(dir string) error {
...
-	// read first line of module file
-	r := bufio.NewReader(modFile)
-	b, _, err := r.ReadLine()
-	if err != nil {
-		log.Fatalf("Error checking module name: %s", err.Error())
+	clientLanguage, jsModulePath := swag.StringValue(c.clientLanguage), swag.StringValue(c.jsModulePath)
+	if swag.BoolValue(c.clientOnly) && swag.BoolValue(c.dynamoOnly) {
+		return fmt.Errorf("Cannot use -dynamo-only and -client-only together")
+	} else if swag.BoolValue(c.clientOnly) {
+		c.generateServer = false
```
taylor-sutton, line 228: Do you think it's important to have this extra verbosity, given that `bool` fields all default to `false` anyway?

main.go
```diff
@@ -334,6 +330,32 @@ func (c *config) setDerivedFields() error {
 		}
 		c.dynamoPath = path.Join(c.goAbsolutePackagePath, *c.relativeDynamoPath)
 	}
+}
 
-	return nil
+func getModulePackagePath(goPath, outputPath string) string {
```
taylor-sutton, line 335: This and `getModulePackageName` didn't change, right? You just moved them?

main.go
```diff
@@ -215,80 +214,80 @@ func prepareDir(dir string) error {
...
-// the function will return github.com/Clever/wag/v6/v2/gen-go
-// Example: if packagePath = github.com/Clever/wag/v6/gen-go and the module name is github.com/Clever/wag/v6
-// the function will return  github.com/Clever/wag/v6/gen-go
-func getModulePackageName(modFile *os.File, outputPath string) string {
-	// read first line of module file
-	r := bufio.NewReader(modFile)
-	b, _, err := r.ReadLine()
-	if err != nil {
-		log.Fatalf("Error checking module name: %s", err.Error())
+	clientLanguage, jsModulePath := swag.StringValue(c.clientLanguage), swag.StringValue(c.jsModulePath)
```
taylor-sutton, line 224: Also a nit: It seems consistent with how you've structured this to factor the block subsequent to this into a `c.setGenerateFlags()` or some such.

taylor-sutton (approved):

main.go
aastein: yes

main.go
aastein: good point. I don't think it's necessary anymore. It was an artifact of the pre-refactor code which set all of this to true.

PR #346, opened 2021-03-03
fix go clients logging circuit breaker for other clients

tnsardesai:
Overview:
Lots of context in https://clever.slack.com/archives/C063LTQP7/p1614742988106900.

TLDR: we run a stream in each client but all point to same map of circuits in memory so every new event irrespective of which circuit is logged by all streams. Ideally we only create one stream but this is fine for the time being.

Rollout:
`make generate` troubleshooting-service and hall-monitor` to see how this actually makes a difference in amount of logs generated. For others services I am going to let this change propagate.

- [x] Run `make build`
- [x] Run `make generate`


taylor-sutton (approved):
LGTM; in order for this change to propagate to `hall-monitor`, though, you'd need to do `make generate` in each of its dependencies, bump their versions, and then pull those new clients into `hall-monitor` by updating its dep version.

As far as just testing, you can go to `hall-monitor` or )`troubleshooting-service` might be easier) and edit the files inside of `vendor/`.

tnsardesai:
> LGTM; in order for this change to propagate to `hall-monitor`, though, you'd need to do `make generate` in each of its dependencies, bump their versions, and then pull those new clients into `hall-monitor` by updating its dep version.
> 
> As far as just testing, you can go to `hall-monitor` or )`troubleshooting-service` might be easier) and edit the files inside of `vendor/`.

Sorry forgot to mention this in the PR description but I tested these changes already by doing exactly that for `troubleshooting-service`.

I didn't realize that `make generate` has to be run in dependencies 💀. I will see how much I can do today

PR #350, opened 2021-03-25
Bump VERSION by a patch

mcab:
- [X] Run `make build`
- [X] Run `make generate`

Wasn't expecting `go-bindata` to do this, but, it happened.


taylor-sutton (approved):

ghirsch1 (approved):
if you don't care about this dependency I think fine not to add it (maybe I'm misunderstanding)

but seems fine 👍

mcab:
> if you don't care about this dependency I think fine not to add it (maybe I'm misunderstanding)
> 
> but seems fine 👍

Just ran `make build`, `make generate`, and it appeared.

I have no idea why it appeared, which is frustrating. :(

ghirsch1:
looks like it's pulling in github.com/tmthrgd/go-bindata instead of github.com/kevinburke/go-bindata 🤷

mcab:
From private convos, I had an older version of `go-bindata` referenced in my path. 🤦

I can at least make a patch to remove the older dependency, and make it explicitly reference `./bin/go-bindata`. :loading:

PR #352, opened 2021-03-25
Fix engineer building mistake

mcab:
- [X] Run `make build`
- [X] Run `make generate`


mcab:
I'm assuming that `Makefile` pointing `PATH` to `$(pwd)/bin:$PATH` accounts for any Makefile invocation.

Makefile
```diff
@@ -73,7 +73,7 @@ release:
 
 install_deps:
 	go mod vendor
-	go build -o bin/go-bindata ./vendor/github.com/kevinburke/go-bindata/go-bindata
-	go build -o bin/mockgen    ./vendor/github.com/golang/mock/mockgen
+	go build -o ./bin/go-bindata ./vendor/github.com/kevinburke/go-bindata/go-bindata
```
taylor-sutton, line 76: Is this a functional difference? Or I'm wondering if in the previous PR, you may not have ran `make install_deps`, and as a result, there was no `./bin/go-bindata` to use, so that even though the `//go:generate` lines DID search `./bin` first, they found nothing and used system `go-bindata` instead.

If that's right, the real difference would be to modify these
https://github.com/Clever/wag/blob/master/hardcoded/gen.go#L3
https://github.com/Clever/wag/blob/master/server/gendb/gendb.go#L18

to `./bin/go-bindata` or similar

rgarcia: I'd prefer not changing go:generate directives to use relative paths since the convention right now is to avoid this and rely on the binary existing in the Makefile PATH

mcab: > Is this a functional difference? Or I'm wondering if in the previous PR, you may not have ran `make install_deps`, and as a result, there was no `./bin/go-bindata` to use, so that even though the `//go:generate` lines DID search `./bin` first, they found nothing and used system `go-bindata` instead.

🤦‍♂️ Oooh, good call.

I'm kind of hesitant to scope all the `go:generate` lines to `./bin/...`, but it does make sense.

taylor-sutton: Perhaps one play is to make two new `Make` targets (one each for `bin/mockgen` and `bin/go-bindata`), have them depend on `install_deps`, and have `make generate` depend on the `bin/*` targets?

mcab: Pretty sure I had very old `/bin` files too, so explicitly checking if `bin/` or `vendor/` exists (to ensure `make install_deps` was run) is also a false sense of confidence.

I'll revert the explicit changes to Makefile and just rebuild in place.

taylor-sutton: (Or perhaps, leaving things as they are but be more diligent catching this in PR review)

mcab: Should be reverted to as before, just removing the accidental changes introduced with 6.4.3.

taylor-sutton (approved):

PR #356, opened 2021-04-15
Binary can be empty

tnsardesai:
Overview:

In https://github.com/Clever/catapult/pull/1374 we want to optionally send a binary file. Today `input.File` is always set to 
`(*models.GradeFile)(&r.Body)` so even if body is empty the File has non-nil `io.ReadCloser`. 

- [x] Run `make build`
- [x] Run `make generate`


samples/gen-go-blog/server/handlers.go
```diff
@@ -160,7 +160,15 @@ func newPostGradeFileForStudentInput(r *http.Request) (*models.PostGradeFileForS
 		input.StudentID = studentIDTmp
 	}
 
-	input.File = (*models.GradeFile)(&r.Body)
+	data, err := ioutil.ReadAll(r.Body)
```
taylor-sutton, line 163: My understanding is that once you call `ReadAll` on `r.Body`, the reader will be exhausted (you can only read it once). So that if even if you store a reference to `r.Body` later on, future `Read` calls will not return anything. 

What is the issue that you're trying to solve here? Is it panicking on the attempted cast? If so, maybe you could try something like
```
gradeFile, ok := (*models.GradeFile)(&r.Body)
if ok {
  input.File = gradeFile
} else {
  input.File = nil
}
```

tnsardesai: you are right. I tested the empty case but not the case where body has some content.

The problem I have is that even if r.Body is empty we are setting `input.File = (*models.GradeFile)(&r.Body)`. As input.File is set to a non-nil value we cant simply do
```
input.File == nil {
 ...
}
```
example use case is [here](https://github.com/Clever/catapult/pull/1374/files#diff-dce67a4c0d3992fa8283dcc21324d69a58e3af9d56c5e82b4de6d580eb4e1771R41-R53)

I thought it made sense to not set this field if body is empty but looks like that is not the pattern we want follow. I can close the PR and use the workaround of reading the field in the app and checking if its empty. 

Ty for catching this

PR #358, opened 2021-04-16
Unpin deps

taylor-sutton:
**This is not asking to merge into master, it is merging into the newly-created wag-7.0-dev branch**. `wag-7.0-dev` is currently [even with master](https://github.com/Clever/wag/tree/wag-7.0-dev)

This is the beginnings of work on v7. This branch is entirely focused on 
- Upgrading dependencies, even if it causes some breakage in `wag`
- Making sure we can build on `go1.16`, and then dropping support for any lower Go versions at the same time.

There's both a lot and not a lot going on in this PR. There are tons of teeny changes to the entirety of the `samples/` directory. But only a small number of manual changes. You can look over the whole diff first ignoring the `samples/`, then spot-checking some `samples`, or you can go commit-by-commit, most commits are just small tweaks and there are a lot of them which serve as logical checkpoints. 

Things I want to call out as particularly important about the manual parts:
- As part of the setup, I'm trying to start spinning up pre-release releases in `.circleci/config.yml` and `VERSION`.

Things that seem particularly important about how `wag` codegen is changing:
- The naming of the Go struct field has changed with regards to capitalization in some cases. For example, I've see that a field named `oncallSlackChannels` in swagger used to become `OncallSLACKChannels` and now becomes `OncallSlackChannels` - this seems to have been a bugfix regarding the initialism `SLA`; fields such as `podIDs` in swagger used to become `PodIds` but now become `PodIDs`; lastly I've seen that a field named `$ne` used to become `NrDollarNe` but now becomes `DollarNe`. The `json` struct tags are the same as they used to be, but this creates a minor breaking change such that 
  1. A server-user of `wag` may need to modify their code to use their the new names
  2. A client-user of `wag` may need to do the same, however because the struct tags are the same, this is only required when pulling in the new version of the client.
- There are some additional `omitempty` tags being added. See below for more details.

Before merging this into `master` as an actual `7.0` release, I plan to write up docs on these and how to preserve existing behavior. For now, I think getting some reps with this is the best way to understand it.

Todo:
- [X] Run `make build`
- [X] Run `make generate`


VERSION
```diff
@@ -1,33 +1,6 @@
-v6.4.4
```
tnsardesai, line 1: Any reason why you had to remove all of these? When we migrated ark to `v3.0.0` we still had the previous release infos. 

If this is temporary add a TODO to re-add versions before release

tnsardesai (approved):
🚢

VERSION
taylor-sutton: I did it because I wanted clean release notes (VERSION is used as the release notes at https://github.com/Clever/wag/releases); I'll add the older versions back under an "Older Versions" header.

PR #375, opened 2021-07-18
Fix client only tracing

tnsardesai:
When I was upgrading `catalog-sync-service` to wag v7 I noticed that client files for the cortex.yaml were updated to include references to tracing.go but that file wasn't generated. This was happening because we were generating tracing files only if we we generating server files.

To reproduce the issue on master
- delete all `gen-*` folders in `samples`
- `make generate`
- `git diff` to see that `samples/gen-go-db-only/tracing/tracing.go` and `samples/gen-go-client-only/tracing/tracing.go` was not generated.

Also notice in this PR the tracing.go file for gen-go-client-only is updated to the latest version!

Todo:

- [x] Run `make build`
- [x] Run `make generate`


taylor-sutton (approved):
Good catch

PR #376, opened 2021-07-23
add close() to js client to release handles at the end of client use

untoldone:
Todo:

- [x] Run `make build`
- [x] Run `make generate`

### DIFF
--- a/VERSION
+++ b/VERSION
@@ -1,4 +1,5 @@
-v7.1.0
+v7.1.1
+- v7.1.1: Add close() to TypeScript clients to release outstanding handles
 - v7.1.0: Properly generate TypeScript types for allOf
 - v7.0.1 bugfix --client-only not generating tracing files
 - v7.0.0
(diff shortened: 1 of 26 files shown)


untoldone:
For context, if WAG client is currently used in the context of a script, the outstanding interval may prevent the process from exiting. Calling `close()` would allow the process to exit.

tnsardesai:
> Failed to run `make build` with:
> 
> ```
> go generate ./hardcoded/
> hardcoded/gen.go:3: running "go-bindata": exec: "go-bindata": executable file not found in $PATH
> make: *** [go-generate] Error 1
> ```

Run `make install_deps` which should take care of setting up `go-bindata`. Also make sure you run `make generate` which will update the sample files

> Also, unsure of how semantic versioning is handled/ didn't know what to rev this to given addition to client public interface
Just update the minor version

untoldone:
@tnsardesai updated code / verified `make build` and `make generate` run successfully. Could you re-review? All changes of note should be in clients/js/genjs.go and the updated VERSION file -- the rest are just re-generated files given the changes.

tnsardesai (approved):

PR #379, opened 2021-07-28
Wag v8 bump openapi

taylor-sutton:
This PR is targeting the wag-8.0-dev branch which will be used for development of wag-8 until we're ready to release it. That branch currently consists JUST of bumping the module suffix to v8 (from v7) and nothing else.

This PR bumps and unpins most go-openapi and go-swagger deps, including account for some breaking changes in how they are used. There are some changes to generated code, which will need to be documented in the eventual v8 release notes (see below). You can find all the non-generated code in the first commit, and the resulting changes to samples in the second commit.

## Go field names
 There are changes to the names of certain fields in the generated Go models. For example, before this PR, a field named `podIDs` used to generate a Go field named `PodIds`, and a field named `oncallSlackChannel` used to generate a field named `OncallSLACKChannel`. After this change, the Go fields are named `PodIDs` and `OncallSlackChannel` (note: I think it picked up on the abbreviation SLA for service level agreement in Slack). 

To get the old behavior, a `wag` user can use the `x-go-name` field. For example:
```yaml
  CaseName:
    type: object
    properties:
      podIDs:
        type: array
        x-go-name: "podIds"
        items:
          type: string
```

## Omitempty changes for types that are alias of slices.

Suppose the following appears in a definitions section:
```yaml
  HasArrayAlias:
    type: object
    properties:
      some_field:
        $ref: "#/definitions/ArrayAlias"

  ArrayAlias:
    type: array
    items:
      type: string
```

In wag v7, the generated model `HasArrayAlias` would have a line like so: 
```
	SomeField ArrayAlias `json:"some_field"`
```

In wag v8, it gains "omitempty": 
```
	SomeField ArrayAlias `json:"some_field,omitempty"`
```

To preserve the old behavior, wecan use `x-omitempty: false` like so:
```

  ArrayAlias:
    x-omitempty: false
    type: array
    items:
      type: string
```

This change ONLY affects *marshaling*, NOT unmarshaling. Before AND after, if a wag controller receives any of the following:
```
{}
{"some_field": null}
{"some_field": []}
```
the resulting `HasArrayAlias` will have `some_field == []string{}` (an empty slice).

It DOES, however, change how marshaling works. Before (in wag 7), `HasArrayAlias{some_field: nil}` marshals into `"some_field":null`, and `HasArrayAlias{some_field: []string{}}` turns into `{"some_field":[]}`. In wag 8, they both turn into `{}` (empty object).

## Todo:
- [X] Run `make build`
- [X] Run `make generate`


rgarcia (approved):
💯

PR #391, opened 2021-08-19
Fix nil pointer panic in retry logic

phenixrizen:
The `RetryPolicy` is an interface which allows us to implement our own retry logic.  The problem is we wanted to handle `context.Timeout` errors. We currently can't because we will panic. We should not assume that there will be a non nil `http.Response`, because in timeout cases there is no response. 

Todo:

- [x] Run `make build`
- [x] Run `make generate`


taylor-sutton (approved):
Broadly it seems like this makes sense. To confirm, the reason you are unhappy with the built-in retry policies is essentially https://github.com/Clever/wag/blob/031b4678fa91f80849f618f9adbdcb018d5c7097/_hardcoded/doer.go#L60
because in this case, case, `err` will be non-nil but we DO actually want to retry. Is that right? In that case, the call to `d.d.Do()` will retry `resp==nil` response and  `err != nil` and we do want to retry, and that's the case that's currently broken?

phenixrizen:
> because in this case, case, `err` will be non-nil but we DO actually want to retry. Is that right? In that case, the call to `d.d.Do()` will retry `resp==nil` response and `err != nil` and we do want to retry, and that's the case that's currently broken?

Correct, we are running into a case where DynamoDB can not scale fast enough if we end up with a large sync in google classroom rostering. When that happens it makes some writes to the API exceed the default wag client timeout. We want to be able catch the `context.Timeout` error and retry.

PR #394, opened 2021-08-24
Quote js field names

taylor-sutton:
Saw in the wild at Clever a case where we have a field with a dash in it in our swagger.yml file. This is valid, since ultimately field names are just strings as far as serialization and deserialization to JSON is concerned. But the generated `index.d.ts` file could not be parsed in some cases, because the field name appeared as 
```
    type LogLine = {
  app-id?: string;
};
```
which is not valid syntax.

This PR will make that appear as 
```
    type LogLine = {
  "app-id"?: string;
};
```
instead. I added a few quick test cases demonstrating this behavior. I did regenerate samples, but none of them were affected by this change.

NOTE: it seems there was a slight change in a comment inside `_hardcoded/doer.go` but it didn't have the updated `hardcoded/hardcoded.go` commited, so there are some tweaks to `samples` due to that.

Todo:

- [X] Run `make build`
- [X] Run `make generate`


clients/js/type_script_test.go
```diff
@@ -57,3 +57,36 @@ func TestGenerateErrorDeclaration(t *testing.T) {
...
+	})
+	t.Run("key has a dash and is required", func(t *testing.T) {
+		assert.Equal(t,
+			`"as-df": any;`,
+			generatePropertyDeclaration("as-df", JSType("any"), true),
+		)
+	})
+	t.Run("key has an emoji", func(t *testing.T) {
+		assert.Equal(t,
+			`"🤩": any;`,
```
tnsardesai, line 88: 🤩

tnsardesai (approved):

PR #400, opened 2021-10-14
Add transaction support to x-db

nikhilb4a:
This PR adds support to the auto-gen dynamo code for transactions. There is an `EnableTransactions` field on the `x-db` object that takes in a list of schemas that the table can write transactions with. The transactions also have conditions that must be met in order for the transaction to succeed in the case of updates. 

One interesting decision is where the definition for which transactions are enabled should live - right now I have it on one of the two schemas involved in the transaction, as that is something that fits with the current design of `x-db` and is a place that felt natural when it comes to using the field (as I set up a table, I may think which other tables I will transact with). That being said, it means that it could end up being unnecessarily set on both tables and it means there are places we need to access the other tables config in order to have the template code follow the current pattern (eg. tests, accessing table name). 

To show how this would be useful, I've created a [PR in messaging-service](https://github.com/Clever/messaging-service/pull/214) that has the changes from our current state to one that uses these wag transactions. I've tested that it builds and the transactions work to fully replace our current approach.

The two situations it replaces are - 
1. Writing a new channel message while updating the corresponding channel object to increment the message count and update the newest message preview
2. Deleting a channel message while updating the corresponding channel object to mark the newest message preview as deleted (if the deleted message is the newest message)

Todo:

This code still has some work to be done including more thorough tests, but Im hoping to get feedback on the approach before continuing down this path.

- [ ] Run `make build`
- [ ] Run `make generate`

### DIFF
--- a/VERSION
+++ b/VERSION
@@ -1,4 +1,6 @@
-v8.0.3
+v8.1.0
+
+v8.1.0: Add dynamodb codegen support for same and cross table transactions
 
 v8.0.3: Fix generated TS types (index.d.ts) when fields might not be valid JS identifiers.
 
@@ -34,7 +36,7 @@ which attributes are set, should not be considered stable.
 - v7.0.0
 
 v7 removes all references to OpenTracing and begins adding OpenTelemetry to replace it. For now,
-OpenTelemetry parts should be considered *unstable*, especially as the OpenTelemetry Go SDKs are
+OpenTelemetry parts should be considered _unstable_, especially as the OpenTelemetry Go SDKs are
 themselves unstable.
 
 Support for the glide package manager for Go is also removed. Previously, wag could update your


server/gendb/interface.go.tmpl
```diff
@@ -60,6 +60,12 @@ type Interface interface {
     Scan{{ $modelName }}sBy{{ $indexName }}(ctx context.Context, input Scan{{ $modelName }}sBy{{ $indexName }}Input, fn func(m *models.{{ $modelName }}, last{{ $modelName }} bool) bool) error
     {{- end }}
     {{- end }}
+    {{- range $transactWithSchemaName := $xdbConfig.EnableTransactions }}
+    {{- $transactWithModelName := pascalize $transactWithSchemaName }}
+    // TransactSave{{ $modelName }}And{{ $transactWithModelName }} saves {{ $modelName }} and {{ $transactWithModelName }} as an atomic transaction.
+    // Use the condition map to indicate pre-transaction condition requirements when updating an item; key = attribute JSON key, use 'x.y' notation for nested properties
```
nbhatia823, line 66: It would be nice if the condition map had type safety (and so maybe was just the actual model itself), but I think we'd lose the ability to distinguish nil values.

server/gendb/table.go.tmpl
```diff
@@ -883,6 +886,87 @@ func (t {{ $modelName }}Table) scan{{ $modelName }}sBy{{ $computedIndexName }}(c
...
+  data1, err := encode{{ $modelName }}(m1)
+  if err != nil {
+    return err
+  }
+
+  m1CondExpr, m1ExprVals, m1ExprNames, err := buildCondExpr(m1ConditionMap)
+  if err != nil {
+    return err
+  }
+
```
nbhatia823, line 903: An [example](https://github.com/Clever/messaging-service/pull/214/files#diff-3672c1b365f00721e784d7201508ceaa44053b758091d5aaf4adf7e191ec89c6R90-R94) of how the conditions can be used

server/gendb/table.go.tmpl
```diff
@@ -883,6 +886,87 @@ func (t {{ $modelName }}Table) scan{{ $modelName }}sBy{{ $computedIndexName }}(c
...
+  return err
+}
+{{- end }}
+
+func buildCondExpr(conditionMap map[string]interface{}) (*string, map[string]*dynamodb.AttributeValue, map[string]*string, error) {
+	var condExpr *string
+	var exprVals map[string]*dynamodb.AttributeValue
+	var exprNames map[string]*string
+	if conditionMap != nil && len(conditionMap) > 0 {
```
taylor-sutton, line 946: Nit: I think you can omit the nil check, if you can `len()` on a slice it just returns 0.

server/gendb/table.go.tmpl
```diff
@@ -883,6 +886,87 @@ func (t {{ $modelName }}Table) scan{{ $modelName }}sBy{{ $computedIndexName }}(c
...
+		c := 0
```
taylor-sutton, line 947: Nit: use a more descriptive variable name.

server/gendb/table.go.tmpl
```diff
@@ -883,6 +886,87 @@ func (t {{ $modelName }}Table) scan{{ $modelName }}sBy{{ $computedIndexName }}(c
...
+          TableName: aws.String(t.name()),
+					Item:      data1,
+					ConditionExpression:       m1CondExpr,
+					ExpressionAttributeValues: m1ExprVals,
+					ExpressionAttributeNames:  m1ExprNames,
+				},
+			},
+			{
+				Put: &dynamodb.Put{
+    			TableName: aws.String(fmt.Sprintf("%%s-{{ $transactWithModelName }}s", t.Prefix)),
```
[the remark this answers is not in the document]
taylor-sutton: Can we add a comment to `t.name()` that if the name format changes, the transaction bit also needs to change?

server/gendb/gendb.go
```diff
@@ -37,6 +37,9 @@ type XDBConfig struct {
 	// DynamoDB configuration.
 	DynamoDB AWSDynamoDBTable
 
+	// EnableTransactions determines which schemas this schema will be able to perform transactions with. It only needs to be set for one per pair.
```
taylor-sutton, line 40: The order of items in a transaction doesn't matter, right? At least in terms of the semantics. (I think maybe if a transaction is going to fail, it might use less read/write capacity on the DB if it fails sooner?)

taylor-sutton: Also, this should work if you put a table to transact with itself?

nbhatia823: 1) The order of the items doesn't matter in regards to semantics and might not even matter when it comes to the underlying db calls, but I'm not certain about that. This line `DynamoDB performs two underlying reads or writes of every item in the transaction: one to prepare the transaction and one to commit the transaction. ` from the docs makes it sound like regardless of the format of the transaction, the underlying read/writes happen in two steps so it seems that the first step is the condition check(s) that all happen regardless of if one fails, and the second write step that may consume the same capacity regardless of failure? Not certain though

2) Correct! With this setup the table can transact with itself

server/gendb/table.go.tmpl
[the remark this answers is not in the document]
nbhatia823: Actually, I realize my reference wasn't accurate. That link is for when the tableName isn't set, which seems to (maybe/probably never?) not be the case? For ex. we pass in the tableNames [here in messaging-service](https://github.com/Clever/messaging-service/blob/master/main.go#L167)

nbhatia823:
Thanks for the review, Taylor! And for our chat the other day to help me more forward. 

> I would also consider accepting the conditions in a more general format - with this API, it's hard to extend this to support anything other than "equals" conditions. Maybe the simplest thing to do is to have the API just accept an expression.ConditionBuilder rather than a map if the plan is not to make some type-safe version using the generated Go model.

This is an interesting point. I've gone back and forth a bit on this. I thought the approach I used might be a good starting point to abstract away the condition building for ease of use if it seemed that most conditions could be satisfied with an "equals" condition (which is the case for both of my use cases), thinking we could maybe expand on in the future if we do see other use cases come up. That being said, since this approach is already not type-safe, maybe it does make sense to go the more general way of accepting an `ConditionBuilder`.

nbhatia823:
Updated the interface to take in a `*expression.ConditionBuilder` instead of an equality condition map. More flexible and still pretty easy to use! Example [here](https://github.com/Clever/messaging-service/blob/passion-demo/controller/channel_messages.go#L92)

taylor-sutton (approved):
I'm excited to see how this gets used! The last thing you need to do before merging is bump the `VERSION` file - which is also used as the release notes, so you can write up some quick notes as well. (First line should stay in the `vX.Y.Z` on the line by itself, then empty line then you can follow the pattern for release notes).

PR #405, opened 2021-11-18
Support TS methods with a single hyphenated header variables

stevenjackson:
Variable names with hyphens won't compile in ts, but they're really
common in header variables.

Convert `x-csrf-token` to `XCSRFToken: string`

If this looks good, I need to backport it to wag7.

### DIFF
--- a/VERSION
+++ b/VERSION
@@ -1,4 +1,6 @@
-v8.1.2
+v8.1.3
+
+v8.1.3: Support TS methods with a single hyphenated header variable
 
 v8.1.2: Improve TS type for close method on js client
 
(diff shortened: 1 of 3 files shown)


venalen (approved):
Code makes sense to me! I'll defer to @rgarcia for any infra-related nuances

nathanleiby (approved):
LGTM.

For backporting, it looks like we'd want to apply this commit to the branch https://github.com/Clever/wag/tree/wag7 and bump the version v7.x https://github.com/Clever/wag/blob/v7.1.1/VERSION as well

taylor-sutton (approved):

PR #409, opened 2022-01-31
update kayvee and otel

rgarcia:
https://clever.atlassian.net/browse/INFRANG-4688

This PR updates to the latest (currently in PR: https://github.com/Clever/kayvee-go/pull/98) version of kayvee, and updates opentelemetry libraries to the latest versions.

Todo:

- [ ] Run `make build`
- [ ] Run `make generate`


rgarcia:
(I recommend going commit-by-commit!)

taylor-sutton (commented):
The changes here seem fine, but I think rollout may be a bit of a hassle for any of our apps that are still set to `latest` wag. In particular, upgrading an app to use the new wag probably requires manually updating the otel dependencies, and changing to the new module path for kayvee-go means any calls in apps' `main()` to [EnableRollups](https://github.com/Clever/kayvee-go/blob/3dcedc88e611b48c93797813d5d0cd2d85033c26/middleware/rollup.go#L22) also need to be changed to import kayvee/v7 and call THAT module's EnableRollups.

taylor-sutton:
Actually the hardest part about rolling this out is probably: if application A uses this and application B uses the old version that depended on the v0.x otel, then neither A can import B, nor B can import A, nor can an application C import both of them.

rgarcia:
yeah... any ideas for how to pull off an upgrade like that? I feel like we have no choice unless we forever don't upgrade otel

taylor-sutton:
> yeah... any ideas for how to pull off an upgrade like that? I feel like we have no choice unless we forever don't upgrade otel

For wag v8, which also had this problem, we just released it and if people had errors due to auto-updating wag, encouraged them to pin to old wag for now. Then some teams have migrated to wag v8 by starting from the bottom of the dependency tree and working up. If A imports B, and B releases a new version using wag v9 (or whatever), the over-the-wire communication between A and B doesn't change, so the old client code used by A can continue to work.

rgarcia:
k, so maybe I'll 
- test this some more with the wag-otel-test app
- write a microplane script to update the otel versions and kayvee imports for the [29](https://github.com/search?p=3&q=org%3AClever+%27WAG_VERSION+%3A%3D+latest%27+filename%3AMakefile&type=Code) repos using `latest` wag
- merge this and open PRs on wag apps
- update template-wag and other template repos with kayvee import

rgarcia:
also maybe we should add some auto-fix code into `wag` itself, like the old code that used to check Godeps for correct versions

PR #412, opened 2022-02-04
Add dynamo batch writes

venalen:
This PR introduces batch write logic into Dynamo. In some of our services, we've been hand-rolling batch write logic (e.g: [gcr-sync-service](https://github.com/Clever/gcr-sync-service/blob/master/batch_write_license_config.patch)) so we may as well add support for it in the autogen.

To generate the batch function we'll need to add the following to the x-db definition:
```
    x-db:
      BatchWriteObject: ArrayOfThingAllowingBatchWrites
```

The code that performs the batch update was taken from the patches that we use in [gcr-sync-service](https://github.com/Clever/gcr-sync-service/blob/master/batch_write_license_config.patch)

- [x] Run `make build`
- [x] Run `make generate`


samples/gen-go-db-custom-path/db/interface.go
```diff
@@ -104,6 +104,23 @@ type Interface interface {
 	// ScanThingsByNameAndCreatedAt runs a scan on the NameAndCreatedAt index.
 	ScanThingsByNameAndCreatedAt(ctx context.Context, input ScanThingsByNameAndCreatedAtInput, fn func(m *models.Thing, lastThing bool) bool) error
 
+	// SaveThingAllowingBatchWrites saves a ThingAllowingBatchWrites to the database.
+	SaveThingAllowingBatchWrites(ctx context.Context, m models.ThingAllowingBatchWrites) error
+	// SaveArrayOfThingAllowingBatchWrites batch saves all items in ArrayOfThingAllowingBatchWrites to the database.
+	SaveArrayOfThingAllowingBatchWrites(ctx context.Context, m models.ArrayOfThingAllowingBatchWrites) error
```
rgarcia, line 110: does this require defining an array type? I guess I was expecting not to have to specify an array type and instead have something like

```yaml
x-db:
  AllowBatchWrites: true
```

```go
SaveArrayOfThingAllowingBatchWrites(ctx context.Context, ms []models.ThingAllowingBatchWrites) 
```

venalen: Unfortunately it does require defining an array. The only reason was because swagger deals with the validation for the models and I couldn't figure out how to get that working without a lot of handrolling. Do you have any suggestions?

rgarcia: Can you expand on what you mean by
>  swagger deals with the validation for the models

venalen: Ah actually looking at it now, I was confusing where the `encode` was coming from within the batch saves itself. Let me take another stab at this where we don't need to define the array!

server/gendb/dynamodb.go.tmpl
```diff
@@ -39,6 +39,8 @@ type Config struct {
   {{- end }}
 }
 
+const maxDynamoDBBatchItems = 25
```
tnsardesai, line 42: Can we add a comment here that this limitation is set by aws. Linking to https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_BatchWriteItem.html

samples/db.yml
```diff
@@ -183,6 +183,48 @@ definitions:
...
+      AllowOverwrites: false
+      AllowPrimaryIndexScan: true
+      AllowBatchWrites: true
+      DynamoDB:
+        KeySchema:
+          - AttributeName: name
+            KeyType: HASH
+          - AttributeName: version
+            KeyType: RANGE
+        GlobalSecondaryIndexes:
```
tnsardesai, line 197: Can we remove the GSI and make this sample simpler? I don't think this block is required for testing `AllowBatchWrites`

tnsardesai (approved):
Please address the 2 comments but otherwise LGTM!

tnsardesai (approved):

PR #415, opened 2022-02-17
DynamoDB support for x-nullable

18vijayb:
**JIRA**: https://clever.atlassian.net/browse/INFRANG-3807

**Overview**: 

Auto-generated code is manually modified due to x-nullable incompatibility. See
https://github.com/Clever/family-invites-service/blob/647b092fc335f2bc9c73a302c03fe8cc8c09888e/Makefile#L34-L37
and https://github.com/Clever/integration-testing-service/blob/master/Makefile#L52

**Testing**:

Ran wag on the two repos above, and the manual changes were implemented by the auto-generated code

Did not fully run make test because of node issue.

- [ ] Run `make build`
- [x] Run `make generate`


server/gendb/table.go.tmpl
```diff
@@ -428,10 +428,10 @@ func (t {{ $modelName}}Table) get{{ $modelName }}sBy{{ $computedIndexName }}(ctx
           {{- if attributeIsPointer $xdbConfig $rangeKey.AttributeName }}
             S: aws.String(toDynamoTimeStringPtr(input.StartingAfter.{{ pascalize $rangeKey.AttributeName }})),
           {{- else }}
-            S: aws.String(toDynamoTimeString(input.StartingAfter.{{ pascalize $rangeKey.AttributeName }})),
+            S: aws.String(toDynamoTimeString({{- if attributeIsXNullable $xdbConfig $rangeKey.AttributeName }}*{{- end}}input.StartingAfter.{{ pascalize $rangeKey.AttributeName }})),
```
tnsardesai, line 431: Can you remove the extra whitespace at the end

server/gendb/table.go.tmpl
```diff
@@ -659,10 +659,10 @@ func (t {{ $modelName}}Table) get{{ $modelName }}sBy{{ $computedIndexName }}(ctx
...
```
tnsardesai, line 662: Same extra space

tnsardesai (commented):
Can you add a test https://github.com/Clever/wag/blob/master/samples/db.yml?

18vijayb:
I ended up rewriting the code because I realized that I had to check if nullable was present in the pointer function simply enough. Let me know what you think @tnsardesai , also added testcases based on those two repos linked above

taylor-sutton (approved):
This is a nice solution!

PR #420, opened 2022-04-01
Filtering dynamo

garrgordon:
Updates various portions of the dynamo db autogenerated code to enable filtering with additional fields when querying the primary index.

1. Adds support to parse AttributeDefinitions in .yml files
2. Adds FIlterValues to input model for dynamoDB queries which allows developers to specify any attribute that is not one of the keys in the primary index as well as values that the attribute should be compared to.
3. Adds FilterExpression to input model to allow developers to build their own filtering expressions

Testing:
Added unit test for filtering in the auto generated tests
Tested filtering locally with il-user-service hitting the MFAConfig table. Worked as expected

Todo:

- [x] Run `make build`
- [x] Run `make generate`


server/gendb/interface.go.tmpl
```diff
@@ -136,6 +153,18 @@ type Get{{ $modelName }}sBy{{ $indexName }}Input struct {
   DisableConsistentRead bool
   // Limit is an optional limit of how many items to evaluate.
 	Limit  *int64
+  {{- if gt $lenNonIndexAttributes 0 }}
+    // Filters is an optional array of filters to apply on various table attributes
```
nbhatia823, line 157: Think this comment should be `// FilterValues..`

server/gendb/interface.go.tmpl
```diff
@@ -108,10 +108,27 @@ type Scan{{ $modelName }}sInput struct {
...
 {{- $rangeKeyAttributes := modelAttributeNamesForKeyType $xdbConfig $xdbConfig.DynamoDB.KeySchema "RANGE" }}
 {{- $lenRangeKeyAttributes := len $rangeKeyAttributes }}
+{{- if gt $lenNonIndexAttributes 0 }}
+    // {{ $modelName }}By{{ $indexName }}FilterableAttribute represents the fields we can apply filters to for queries on this index
+    type {{ $modelName }}By{{ $indexName }}FilterableAttribute string
+    {{- range $attributeName := $nonIndexAttributeNames }}
+      const {{ $modelName }}{{ pascalize $attributeName }} {{ $modelName }}By{{ $indexName }}FilterableAttribute = "{{ $attributeName }}"
+    {{- end }}
+
+    // {{ $modelName }}By{{ $indexName }}Filter represents a filter on a particular field to be included in the query
```
nbhatia823, line 124: Same as below, should be `...FilterValues`

server/gendb/interface.go.tmpl
```diff
@@ -108,10 +108,27 @@ type Scan{{ $modelName }}sInput struct {
...
+    type {{ $modelName }}By{{ $indexName }}FilterValues struct {
+      // AttributeName is the attibute we are attempting to apply the filter to
+      AttributeName {{ $modelName }}By{{ $indexName }}FilterableAttribute
+      // AttributeValues is an optional parameter to be used when we want to compare the attibute to a single value or multiple values
+      AttributeValues []interface{}
```
nbhatia823, line 129: What would it mean to not have `AttributeValues` set? Wondering since the comment says it is optional

server/gendb/interface.go.tmpl
```diff
@@ -136,6 +153,18 @@ type Get{{ $modelName }}sBy{{ $indexName }}Input struct {
...
+    // Filters is an optional array of filters to apply on various table attributes
+    FilterValues []{{ $modelName }}By{{ $indexName }}FilterValues
+    // FilterExpression is the filter expression to be applied to our fitlered attributes 
+    // when referencing an attribute use #ATTRIBUTE_NAME
+    // ex: if the attribute is called "created_at" in its wag definition use #CREATED_AT
+    // when referencing one of the given values use :{attribute_name}_value0, :{attribute_name}_value1, etc.
+    // ex: if the attribute is called "created_at" in its wag definition use :created_at_value0, created_at_value1, etc.
+    // see https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/Query.html#Query.KeyConditionExpressions
+    // for guidance on building expressions
+    FilterExpression  string
```
nbhatia823, line 166: Is there any validation to ensure that if FilterValues are provided that a FilterExpression is also provided? If not, what would it mean if only one was provided

server/gendb/interface.go.tmpl
```diff
@@ -136,6 +153,18 @@ type Get{{ $modelName }}sBy{{ $indexName }}Input struct {
...
   // Limit is an optional limit of how many items to evaluate.
 	Limit  *int64
+  {{- if gt $lenNonIndexAttributes 0 }}
+    // Filters is an optional array of filters to apply on various table attributes
+    FilterValues []{{ $modelName }}By{{ $indexName }}FilterValues
+    // FilterExpression is the filter expression to be applied to our fitlered attributes 
+    // when referencing an attribute use #ATTRIBUTE_NAME
+    // ex: if the attribute is called "created_at" in its wag definition use #CREATED_AT
+    // when referencing one of the given values use :{attribute_name}_value0, :{attribute_name}_value1, etc.
+    // ex: if the attribute is called "created_at" in its wag definition use :created_at_value0, created_at_value1, etc.
```
nbhatia823, line 163: How come here we are using the lower case name but on line 160 we use the upper case? I tried to look at the link below but didn't find the answer

server/gendb/table.go.tmpl
```diff
@@ -513,12 +553,26 @@ func (t {{ $modelName}}Table) get{{ $modelName }}sBy{{ $computedIndexName }}(ctx
...
+  {{- end }}
+
   totalRecordsProcessed := int64(0)
   var pageFnErr error
 	pageFn := func(queryOutput *dynamodb.QueryOutput, lastPage bool) bool {
-		if len(queryOutput.Items) == 0 {
-			return false
-		}
+		{{- if gt $lenNonIndexAttributes 0 }}
+      // Only assume an empty page means no more results if there are no filters applied
```
nbhatia823, line 567: Sorry I'm not sure I understand why this is the case, could you explain?

server/gendb/tests.go.tmpl
```diff
@@ -538,7 +649,11 @@ func Scan{{ $modelName }}s(d db.Interface, t *testing.T) func(t *testing.T) {
             // in secondary indexes, since dynamodb doesn't support
             // empty strings:
             {{- range $attr := $nonPKSecondaryStringProperties }}
-              {{ pascalize $attr }}: firstItem.{{ pascalize $attr }},
+              {{- if attributeIsPointer $xdbConfig $attr }}
+                {{ pascalize $attr }}: aws.String("{{ $attr }}"),
+              {{- else }}
+                {{ pascalize $attr }}: "{{ $attr }}",
```
nbhatia823, line 655: Was the `firstItem.` intentionally dropped?

nbhatia823 (commented):
Left some questions, mostly for my understanding, since I'm not too familiar with the intended use-case. Mostly LGTM, though!

server/gendb/interface.go.tmpl
garrgordon: Most likely it would mean that the filter is checking for the existence of the field in the record.  In that case we wouldn't need any reference to a value just the field name itself

server/gendb/table.go.tmpl
garrgordon: When we query a page of records we first fetch them and then apply filters returning whatever is not filtered out as a page of results.  This is true even if we filter out all the records.  It is possible then that we get an empty result set when querying for a page when there are still more results in the table for the given hash/range key.  Previously because filters could not be provided we assumed an empty page meant there were no more results in the table for the given keys

server/gendb/interface.go.tmpl
garrgordon: No validation that throws an error but we have this at line 557 `if len(input.FilterValues) > 0 && input.FilterExpression != ""`. if both are not provided then no filters will be applied

server/gendb/interface.go.tmpl
garrgordon: it's just a naming convention that I've chosen to be similar to the naming convention used already for the hash/range keys.  it doesn't necessarily have to be this way.  See https://github.com/Clever/wag/blob/e2e3ba730ab6ec7243fbbc5b1a860c70eb973aa3/samples/gen-go-db/server/db/dynamodb/thing.go#L324 for an example of how its already used for keys

server/gendb/interface.go.tmpl
nbhatia823: Could you add a comment above that line explaining why we are doing that check? Just so it's clear that both fields are needed to filter, and if one is missing the other is ignored

server/gendb/table.go.tmpl
nbhatia823: Ah okay, that makes sense. could you clarify that in the comment?

Also (doing my best to parse this template code, which is somewhat hard haha) - in the case where we have a filter expression provided and `len(queryOutput.Items) == 0`, should we be returning `!lastPage`. I _think_ right now we are just returning `true` at line 595, after going through some code that wont do anything since there are no `queryOutput.items`

nbhatia823 (approved):
LGTM! Few comments about updating comments for clarity, and one question about the return value within the paging function when there are no items retrieved due to the filter

PR #422, opened 2022-04-18
Fix dynamodb filtering ordering

tnsardesai:
When I was working on a different ticket I noticed that db code was getting updated even with no change! You can recreate this issue by just running make generate in this repo multiple times

 This PR fixes that issue by sorting the list

- [x] Run `make build`
- [x] Run `make generate`


taylor-sutton (approved):
Thanks for fixing this! It annoyed me too.

Issue #328, opened 2021-02-05
Do not rely on GOPATH


aastein:
Go is deprecating GOPATH in favor of modules.
https://blog.golang.org/modules2019

We need to somehow use golang.org/x/tools/go/packages to determine file paths

PR #423, opened 2022-04-21
Remove GOPATH (breaking change for using wag outside of GOPATH)

taylor-sutton:
This is a merge request into the `v9.x` dev branch, currently even with `master` but this will be where it starts to diverge.

This PR is split into two commits.

The first commit is fairly mechanical: the idea is to reduce any remaining uses of GOPATH to the top-level, rather than using it deeper into the generator. The deeper usage of GOPATH that needed to be eliminated is in the `Generator` type in our `swagger` package. Previously this struct was receiving a path that was interpreted relative to GOPATH, now it's receiving an absolute path.

In addition, I tried to simplify a bit by removing the number of `*path` variables. For example, in practice, `modelsPath` was always `{basePath}/models`; yet at the same time, the assumption that models lived in that directory was hardcoded in other places, so that if we just changed `modelsPath`, things would actually break.

The second commit included a **BREAKING CHANGE** ⚠️ 🔴 . This change was necessary to completely remove any dependence on GOPATH. Now the -output-path, if it's relative, is interpreted relative to current working directory. Previously, we would combine GOPATH with the module path in go.mod to guesstimate where to interpret it relative to.

This change is only sorta kinda breaking. If you were using `wag` inside `${GOPATH}/src/path/to/package` already, nothing changes for you. Using wag outside such a directory... was already kinda broken. So while this is a breaking change, it's really fixing things. Previously, if I ran `wag` in `~/someDir` (i.e. outside of GOPATH), `wag` would put the generated files inside of GOPATH, which is basically never what you want.

## Testing

Running `make test` not only passes the test, it also results in literally zero diffs inside samples, meaning that this is not changing anything in the use cases we are about.

I also tried running wag outside of GOPATH... and it actually worked! It produced generated files inside the working directory instead of randomly far away.

## Merge Checklist

- [X] Run `make build`
- [X] Run `make generate`
- [ ] Update the current version in the `/VERSION` file - N/A, not trying to release yet.

Closes https://github.com/Clever/wag/issues/328


andruwm (approved):
Nice cleanup!

PR #437, opened 2022-07-22
Wag nightly build

ChrisScotMartin:
Draft PR:

Most of this work revolves around solving dependency problems, it does so through further abstraction and breaking things into modules. 

[This doc](https://docs.google.com/document/d/184-Gkg_BTrw_t7mhu175eI_O-JvIPZMKCjYywWWZLiU/edit?usp=sharing) details the work and thoughts behind this draft PR, I figured it's big enough to need a google doc rather than just a story here.


models/genmodels.go
```diff
@@ -45,18 +45,56 @@ func Generate(packagePath string, s spec.Swagger) error {
...
+
+	`
+	_, err = f.WriteString(modFileString)
+
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
```
ChrisScotMartin, line 94: Still want to make this a go mod init followed by a bunch of go gets. 
Or even just an empty go.mod with the go get calls. Here and in client/gengo.go

loggers/printlogger/printLogger.go
```diff
@@ -0,0 +1,49 @@
...
+		return 2
+	case "Info":
+		return 3
+	case "Debug":
+		return 4
+	case "Trace":
+		return 5
+	}
+	return -1
+}
```
ChrisScotMartin, line 49: Will reverse these to match kayvee, was based off of the levels in https://github.com/go-logr/logr

main.go
```diff
@@ -95,31 +90,31 @@ func main() {
 	}
 
 	if conf.generateGoModels {
-		if err := generateGoModels(conf.modelsPath, conf.goPackagePath, swaggerSpec); err != nil {
+		if err := generateGoModels(*conf.goPackageName, conf.goAbsolutePackagePath, swaggerSpec); err != nil {
```
ChrisScotMartin, line 93: These changes were part of getting the name of the package that the model is being generated for accessible when we generate the code.  Necessary for placing the go.mod file and for later versioning of the models module.

_hardcoded/doer.go
```diff
@@ -30,7 +29,8 @@ type opNameCtx struct{}
 type baseDoer struct{}
 
 func (d baseDoer) Do(c *http.Client, r *http.Request) (*http.Response, error) {
-	return ctxhttp.Do(r.Context(), c, r)
+	return c.Do(r)
```
ChrisScotMartin, line 32: Getting rid of severely outdated ctxhttp dependency.

tracing/tracing.go
```diff
@@ -0,0 +1,122 @@
+package tracing
```
ChrisScotMartin, line 1: This is a tracing module meant to be imported by an app that also imported a wag client. 

It allows for instrumenting tracing using our defaults.

clients/go/gengo.go
```diff
@@ -72,38 +85,196 @@ type WagClient struct {
...
+	r, _ := resource.Merge(
+		resource.Default(),
+		resource.NewWithAttributes(
+			semconv.SchemaURL,
+			semconv.ServiceNameKey.String("dapple"),
+			semconv.ServiceVersionKey.String("1.11.0"),
+		),
+	)
+	return r
+}
```
ChrisScotMartin, line 180: Need to remove this as I don't want the semconv dep. Easy to pass {{.ServiceName}} and {{.Version}} to something else (wag/tracing) that then generates this resource. Will make that change today. It's just more passing back and forth that I don't like/can't get away from.

clients/go/gengo.go
```diff
@@ -72,38 +85,196 @@ type WagClient struct {
...
+	logger       waglogger.WagClientLogger
+	instrumentor Instrumentor
+	exporter     sdktrace.SpanExporter
+}
+
+type Option interface {
+	apply(*options)
+}
+
+//Logger
```
rgarcia, line 106: nit: this comment (and others like it that aren't attached to any code) don't really add to my understanding of the file and I think they can be removed

clients/go/gengo.go
```diff
@@ -72,38 +85,196 @@ type WagClient struct {
...
+
+type roundTripperOption struct {
+	rt http.RoundTripper
+}
+
+func (t roundTripperOption) apply(opts *options) {
+	opts.transport = t.rt
+}
+
+//WithRoundTripper allows you to pass in intrumented/custom roundtrippers which will then wrap the
```
rgarcia, line 131: nit but most of our Go code comments have a space between the `//` and the start of the comment, i.e.
```suggestion
// WithRoundTripper allows you to pass in intrumented/custom roundtrippers which will then wrap the
```

loggers/printlogger/printLogger.go
```diff
@@ -0,0 +1,49 @@
...
+import (
+	"encoding/json"
+	"fmt"
+
+	logger "github.com/Clever/wag/loggers/waglogger"
+)
+
+//NewLogger creates a logger for id that produces logs at and below the indicated level.
+//Level indicated the level at and below which logs are created.
+func NewLogger(id string, level string) logger.WagClientLogger {
```
rgarcia, line 12: this can return PrintlnLogger and then this package has no dependencies

rgarcia (commented):
did a first pass. It's looking good! Haven't looked at the tracing stuff in depth

loggers/waglogger/WagClientLogger.go
```diff
@@ -0,0 +1,9 @@
+package logger
+
+//WagClientLogger provides a minimal interface for a Wag Logger
+type WagClientLogger interface {
```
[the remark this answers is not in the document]
ChrisScotMartin: I'm up for that change.

_hardcoded/middleware.go
```diff
@@ -5,7 +5,7 @@ import (
 	"net/http"
 	"runtime/debug"
 
-	"gopkg.in/Clever/kayvee-go.v6/logger"
+	"github.com/Clever/kayvee-go/v7/logger"
```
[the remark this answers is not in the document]
ChrisScotMartin: Yeah, do we want to do that? Or is removing the client dep on kayvee good enough?

clients/go/gengo.go
ChrisScotMartin: Gotcha, I'll remove them later, they were mostly for MY understanding/organization of the code.

loggers/printlogger/printLogger.go
ChrisScotMartin: Oooh, nice catch. Thanks!

_hardcoded/middleware.go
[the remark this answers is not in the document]
rgarcia: kayvee updates pretty frequently so I think removing the kayvee dep on the server would be beneficial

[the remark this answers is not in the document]
ChrisScotMartin: My thinking: I don't see a case for ever importing a wag server into a seperate app. We are always going to want logging on the server side, and we currently use kayvee for that. So any generated server is going to be part of the app that generated it (ie. dapple) and will need to depend on kayvee anyway. We can abstract away to a WagServerLogger but at some point I'll still need to feed it a kayvee logger, and that point will need to be in the generated code. 

Unless I'm missing something.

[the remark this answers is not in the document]
rgarcia: as soon as you publish a version of wag that generates server code that uses kayvee v7, then every wag app now needs to upgrade to v7, which is annoying and ties wag upgrades to kayvee upgrades

if instead wag generates server code that does not mention kayvee, then all of our wag apps can continue to use whatever version of kayvee they were using before

hardcoded/hardcoded.go
```diff
@@ -1,17 +1,16 @@
-// Code generated by go-bindata. DO NOT EDIT.
+// Code generated for package hardcoded by go-bindata DO NOT EDIT. (@generated)
```
taylor-sutton, line 1: Did the version of `go-bindata` change? I don't see a corresponding change in `go.mod`.

clients/go/gengo.go
```diff
@@ -323,15 +323,15 @@ func New(ctx context.Context, basePath string, opts ...Option) *WagClient {
 
 // NewFromDiscovery creates a client from the discovery environment variables. This method requires
 // the three env vars: SERVICE_{{.FormattedServiceName}}_HTTP_(HOST/PORT/PROTO) to be set. Otherwise it returns an error.
-func NewFromDiscovery() (*WagClient, error) {
+func NewFromDiscovery(opts ...Option) (*WagClient, error) {
```
taylor-sutton, line 326: This change makes me wonder if there's a different default that might be better or safer. Thoughts?

ChrisScotMartin: So I spent a while Friday experimenting with that. 

And I know it's something you mentioned here:

```
POSSIBLE ANNOYANCE: Doing things this way - where the client default is not our Clever best practice, but you have to call some library whenever you create a wag client to get it set up with tracing - is okay but not ideal, since it makes it easy to lose instrumentation by accident. Maybe there’s some way to have it all - like call some global “set default wag client options” method then whenever clients are constructed, they’ll pull from that first? Seems tricky to make work without having the client depend on e.g. particular versions of github.com/Clever/wag or the library that includes this thing, etc, but maybe.
```

The actual setup isn't "too" ugly and looks like:
```go 
	dacService, err := dacServiceClient.NewFromDiscovery(
		dacServiceClient.WithLogger(logger.NewConcreteLogger("dac-service-client")),
		dacServiceClient.WithInstrumentor(tracing.InstrumentedTransport),
		dacServiceClient.WithExporter(exporter),
	)
```

The trick is what we REALLY want to do is basically be able to pass in a client "package" to a function and then have it call that package's functions with our options. And there just isn't a way to do that in Go. 

```go 
dacService, err := someConfigPackage.AutoInstrument(dacServiceClient);
```

_hardcoded/tracing.go
```diff
@@ -10,119 +10,149 @@ import (
...
-	exporter, err := otlp.NewExporter(
-		ctx,
-		driver,
+	otlpClient := otlptracegrpc.NewClient(
+		otlptracegrpc.WithReconnectionPeriod(15*time.Second),
+		otlptracegrpc.WithEndpoint(addr),
+		otlptracegrpc.WithInsecure(),
 	)
+	spanExporter, err := otlptrace.New(ctx, otlpClient)
+	// spanExporter, err := jaeger.New(jaeger.WithCollectorEndpoint(jaeger.WithEndpoint("http://localhost:14268/api/traces")))
```
taylor-sutton, line 46: When would one uncomment this? Maybe add an extra note along the lines of "to test locally, you can use Jaeger and use this line instead of the previous"

_hardcoded/tracing.go
```diff
@@ -10,119 +10,149 @@ import (
...
 		// These maximums are to guard against something going wrong and sending a ton of data unexpectedly
 		sdktrace.WithSpanLimits(sdktrace.SpanLimits{
 			AttributeCountLimit: 100,
 			EventCountLimit:     100,
 			LinkCountLimit:      100,
 		}),
-		sdktrace.WithSyncer(exporter),
+
+		// Batcher is more efficient, switch to it after testing
+		// sdktrace.WithSyncer(exporter),
```
taylor-sutton, line 88:
```suggestion
```

_hardcoded/tracing.go
```diff
@@ -10,119 +10,149 @@ import (
...
+		samplingProbability = samplingProbabilityFromEnv
+	}
 
-func newTracerProvider(exporter sdktrace.SpanExporter, samplingProbability float64) *sdktrace.TracerProvider {
-	return sdktrace.NewTracerProvider(
+	tp := sdktrace.NewTracerProvider(
 		// We use the default ID generator. In order for sampling to work (at least with this sampler)
 		// the ID generator must generate trace IDs uniformly at random from the entire space of uint64.
 		// For example, the default x-ray ID generator does not do this.
+		// sdktrace.WithSampler(sdktrace.TraceIDRatioBased()),
```
taylor-sutton, line 78:
```suggestion
```

clients/go/gengo.go
```diff
@@ -73,54 +90,246 @@ type WagClient struct {
...
+	}
+
+	for _, o := range opts {
+		o.apply(&options)
+	}
+
+
+	samplingProbability := 1.0 // TODO: Hard setting this to one for now, because right now 
+	// it is essentially ignored as the sidecar is determining the sample rate it forwards on to DD.
+	// samplingProbability := determineSampling()
```
taylor-sutton, line 293: Remove the TODO and just explain that the preferred approach is to sample locally in the sidecar?

server/gendb/dynamodb.go.tmpl
```diff
@@ -4,8 +4,7 @@ import (
   "context"
   "errors"
   "time"
-
```
taylor-sutton, line 7: I think this whitespace deletion was accidental?

server/gendb/gendb.go
```diff
@@ -137,8 +139,21 @@ func findCompositeAttribute(config XDBConfig, attributeName string) *CompositeAt
 
 }
 
+func extractModuleNameAndVersionSuffix(packageName, outputPath string) (moduleName, versionSuffix string) {
+	vregex, err := regexp.Compile("/v[0-9]$|/v[0-9][0-9]")
```
taylor-sutton, line 143: Any reason to only support two-digit major versions?

server/genserver.go
```diff
@@ -186,11 +197,24 @@ func jsonMarshalNoError(i interface{}) string {
 {{end}}
 `
 
-func generateHandlers(packageName, packagePath string, s *spec.Swagger, paths *spec.Paths) error {
+func extractModuleNameAndVersionSuffix(packageName, outputPath string) (moduleName, versionSuffix string) {
```
taylor-sutton, line 200: This function seems to appear exactly in 3 or more places?

taylor-sutton (commented):
Can't seem to comment on a binary file but you can also remove the `wag` binary at the top level of the repo

clients/go/gengo.go
```diff
@@ -186,7 +395,7 @@ func (c *WagClient) SetTimeout(timeout time.Duration){
 
 // SetTransport sets the http transport used by the client.
 func (c *WagClient) SetTransport(t http.RoundTripper){
-	c.client.Transport = tracing.NewTransport(t, opNameCtx{})
+	// c.client.Transport = tracing.NewTransport(t, opNameCtx{})
```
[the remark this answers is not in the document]
ChrisScotMartin: It doesn't and is now removed. To make this function actually do something we would need to import the otel tracing library in the client which we are avoiding doing. Instead this is handled by setting the transport and instrumentor at client creation.

server/genserver.go
ChrisScotMartin: Yep, should have written a TODO for that one. I moved it into utils.

server/gendb/gendb.go
ChrisScotMartin: That was a compromise that I asked you about pretty early on, it's possible to write this to support larger version numbers but it adds complexity and reduces readability in this line. As of now I believe we have exactly one wag service that has hit 2 digit version numbers. I don't expect us to get to anything have a three digit version number.

taylor-sutton (approved):

PR #438, opened 2022-07-28
recover response body from unknown response code responses

rgarcia:
https://clever.atlassian.net/browse/INFRANG-4969

@taylor-sutton can you take a look and see if this is along the lines of what you were thinking in terms of a solution to this?


samples/gen-go/client/client.go
```diff
@@ -1079,6 +1081,11 @@ func (c *WagClient) doGetBookByIDRequest(ctx context.Context, req *http.Request,
 		return nil, &output
 
 	default:
+		var body interface{}
+		bs, err := ioutil.ReadAll(resp.Body)
+		if err == nil && json.Unmarshal(bs, &body) == nil {
```
taylor-sutton, line 1086: This is the kind of change I was thinking of, but I'll note:
- it seems like `body` is unused? I'm not sure if we want to special case if the body is JSON vs if it isn't - if the response is an unexpected status code we don't want to make assumptions about the body structure.
- It would be good to include the unknown status code in the error.

I also am not sure either way if it makes sense to create a `models.UnknownResponseCode` model to encapsulate this, or just use `fmt.Errorf`/`errors.New`

rgarcia: ah ok. I got the sense that part of the solution would be some kind of inference about whether the server had created a new response code, and that the presence of a JSON body would be indicative of that

I like the idea of a new error type so will noodle on that

rgarcia: @taylor-sutton just made changes to use `models.UnknownResponse` let me know what you think

models/genmodels.go
```diff
@@ -359,6 +359,14 @@ func (o %s) Error() string {
 `, errorType))
 	}
 
+	// default unknown response type is also an error type
+	buf.WriteString(`
+	func (u UnknownResponse) Error() string {
+		return fmt.Sprintf("unknown response with status: %%d body: %%s", u.StatusCode, u.Body)
```
taylor-sutton, line 365: Was the code in samples generated with this? I'm not sure I understand the doubled-`%` since this isn't being `fmt.Printf`'d.

models/genmodels.go
```diff
@@ -308,7 +308,7 @@ var queryParamStr = `
 func generateOutputs(packagePath string, s spec.Swagger) error {
 	g := swagger.Generator{PackagePath: packagePath}
 
-	g.Printf("package models\n\n")
+	g.Printf("package models\nimport \"fmt\"\n")
```
taylor-sutton, line 311: This should have doubled `\n` to be properly formatted.

taylor-sutton (commented):
This looks like a good path to me. I commented with two small nits. Also it might be worth a check to make it an error if the user defined an `UnknownResponse` in their swagger spec?

models/genmodels.go
rgarcia: 👍

models/genmodels.go
rgarcia: Yes, this code generates what's in the samples directory. I don't quite know what's going on, but when I didn't have the double %s it showed up as `MISSING` and the double %s fixed it.

PR #443, opened 2022-09-01
Wag nightly build

ChrisScotMartin:
Minor fixes pointed out in [mongo-user-generator PR](https://github.com/Clever/mongo-user-generator/pull/16)


_hardcoded/tracing.go
```diff
@@ -84,10 +84,10 @@ func newTracerProvider(exporter sdktrace.SpanExporter, resource *resource.Resour
 			EventCountLimit:     100,
 			LinkCountLimit:      100,
 		}),
-		//Batcher is more efficient, switch to it after testing
+
+		// Batcher is more efficient, switch to it after testing
 		sdktrace.WithSyncer(exporter),
-		//sdktrace.WithBatcher(exporter),
-		//Have to figure out how I'm going to generate this resource first.
+		// sdktrace.WithBatcher(exporter),
```
andruwm, line 90: Does this need to get changed back?

_hardcoded/tracing.go
```diff
@@ -160,9 +160,9 @@ func MuxServerMiddleware(serviceName string) func(http.Handler) http.Handler {
 // Used for setting up tracer provider
 func newResource() *resource.Resource {
 	var appName string
-	if os.Getenv("_POD_ID") != "" {
+	if os.Getenv("_APP_NAME") != "" {
```
andruwm, line 163: I know you were fixing a bug here so optional change. This can be simplified as
```go
if an := os.Getenv("_APP_NAME"); an != "" {
    appName = an
}
```

clients/go/gengo.go
```diff
@@ -243,7 +243,7 @@ func doNothing(baseTransport http.RoundTripper, spanNameCtxValue interface{}, tp
 func determineSampling() (samplingProbability float64, err error) {
 
 		// If we're running locally, then turn off sampling. Otherwise sample
-		// 1% or whatever TRACING_SAMPLING_PROBABILITY specifies.
+		// 1%% or whatever TRACING_SAMPLING_PROBABILITY specifies.
```
andruwm, line 246: Does this get parsed by package `fmt`?

andruwm: Ah I just read the previous PR, so I understand these changes. Still curious about what parses this string?

ChrisScotMartin: It's actually parsed by template. 

https://github.com/Clever/wag/blob/7904b872303587fb952883c24f27aca2e8d1dc39/clients/go/gengo.go#L450-L451

andruwm: Interesting. I didn't realize template uses replacement the way `fmt` does.

_hardcoded/tracing.go
ChrisScotMartin: I think it might be worth doing at this point. FYI, we've had it set to WithSyncer in prod this whole time. My only worry is that while we are testing on low use services it might take a while for it to form a big enough batch to send to the collector, making testing harder.

_hardcoded/tracing.go
ChrisScotMartin: That actually makes sense, but I want to leave the current formatting because I'm thinking I'll want to add more env vars to the resource later. Right now I'm doing the bare minimum.

clients/go/gengo.go
ChrisScotMartin: Neither did I until I saw (MISSING) showing up in my comments.

ChrisScotMartin:
Sorry, just realized I hadn't merged this before iterating again. All I did was add baggage propagation.

_hardcoded/tracing.go
andruwm: Ah I see, if you're only on low use services for now and this is intended then I'm not worried about leaving this in. The extra context helps. Thanks!

andruwm (approved):

PR #445, opened 2022-09-16
Infrang 5044

ChrisScotMartin:
https://clever.atlassian.net/browse/INFRANG-5044

Some of the wag v9 changes resulted in wag additional modules being added to the wag repo as “sub-modules” that wag clients import. 

To ease the transition it would be helpful to commit these modules to the current master BEFORE finalizing the rest of the wag v9 changes as this would allow clients created with wag v9 to import these modules with no code changes / hard coding that would have to eventually be removed.

Since these are not part of the main wag module they should in no way affect the current wag. 

(Running make build/generate updated go/bindata as well, I can always revert that change but it doesn't seem to cause an issue)

- [x] Run `make build`
- [x] Run `make generate`


logging/wagclientlogger/wagclientlogger.go
```diff
@@ -0,0 +1,23 @@
+package wagclientlogger
+
+//WagClientLogger provides a minimal interface for a Wag Client Logger
+type WagClientLogger interface {
```
andruwm, line 4: Out of curiosity, is there a spec for the new logging interfaces (both client and server)? I'd be interested in reading it.

tracing/tracing.go
```diff
@@ -0,0 +1,121 @@
...
+type options struct {
+	address string
+}
+
+// WithAddress takes an address in the form of Host:Port
+func WithAddress(addr string) Option {
+	return addrOption{address: addr}
+}
+
+type addrOption struct {
```
andruwm, line 34: Not a required change for this PR: gRPC demonstrates a good pattern for generating options without needing to define a struct for each one – https://github.com/grpc/grpc-go/blob/master/dialoptions.go#L105

tracing/tracing.go
```diff
@@ -0,0 +1,121 @@
...
+
+func (o addrOption) apply(opts *options) {
+	opts.address = o.address
+}
+
+// OtlpGrpcExporter uses the otlptracegrpc modules and the otlptrace module to produce a new exporter at our default addr
+func OtlpGrpcExporter(ctx context.Context, opts ...Option) sdktrace.SpanExporter {
+
+	DefaultCollectorHost := "localhost"
+	var defaultCollectorPort uint16 = 4317
```
andruwm, line 46: nit: magic numbers and strings should be consts

tracing/tracing.go
```diff
@@ -0,0 +1,121 @@
...
+
+	otlpClient := otlptracegrpc.NewClient(
+		otlptracegrpc.WithEndpoint(addr), //Not strictly needed as we use the defaults
+		otlptracegrpc.WithReconnectionPeriod(15*time.Second),
+		otlptracegrpc.WithInsecure(),
+	)
+
+	spanExporter, err := otlptrace.New(ctx, otlpClient)
+	if err != nil {
+		log.Fatal(err)
```
andruwm, line 65: Fatal logging in a lib is generally an anti-pattern. What about calling OtlpGrpcExporter makes it hard to handle a returned error?

tracing/tracing.go
```diff
@@ -0,0 +1,121 @@
...
+	}
+	return spanExporter
+
+}
+
+func JaegerExporter() (spanExporter sdktrace.SpanExporter) {
+	fmt.Println("Creating Jaeger Exporter")
+	spanExporter, err := jaeger.New(jaeger.WithCollectorEndpoint(jaeger.WithEndpoint("http://localhost:14268/api/traces")))
+	if err != nil {
+		log.Fatal("Error creating Jaeger Exporter")
```
andruwm, line 78: Can this return an error instead?

tracing/tracing.go
```diff
@@ -0,0 +1,121 @@
...
+		//Is doing a fatal error here too risky? No easy way to bubble up errors from here to the app using this.
+		//without making each of the WithXOption() takes an error as an arg as well.
+		return nil
+	}
+	return spanExporter
+
+}
+
+func JaegerExporter() (spanExporter sdktrace.SpanExporter) {
+	fmt.Println("Creating Jaeger Exporter")
```
andruwm, line 75: This should probably be done by the caller if needed.

tracing/tracing.go
```diff
@@ -0,0 +1,121 @@
...
+	spanExporter, err := jaeger.New(jaeger.WithCollectorEndpoint(jaeger.WithEndpoint("http://localhost:14268/api/traces")))
```
andruwm, line 76: Should this be hardcoded? const with explanation if so.

tracing/tracing.go
ChrisScotMartin: Oooh, a little obfuscated...but then again so is making a ton of structs and this is a lot easier to maintain. Will keep in mind for the future.

logging/wagclientlogger/wagclientlogger.go
ChrisScotMartin: It was briefly mentioned in the original wagv9 spec by Taylor. There is no new logging interface for the server (but it is a change we want to eventually make). The idea was just a minimal wrapper that can be easily applied to any logger but similar to kayvee-go. We plan on using kayvee-go for the foreseeable future but don't want to have to import it into the client. 

It will eventually be totally obsoleted by [this](https://github.com/golang/go/discussions/54763) hopefully.

tracing/tracing.go
ChrisScotMartin: Will change.

tracing/tracing.go
ChrisScotMartin: Honestly, all of the jaeger stuff can be removed. We don't use it in prod/dev it just makes local development easier as I don't have to create an opentelemetry sidecar and can just launch the jaeger-all-in-one to collect/view traces. 

I'll remove and then add back later as a separate module that can be imported if desired.

tracing/tracing.go
ChrisScotMartin: OtlpGrpcExporter is passed in as an WithExporter() option to the new() function of a wag client. 

Example:
`func createClient() (client *wot2.WagClient) {
	ctx := context.Background()
	client = wot2.New(ctx,
		"https://csm--wag-otel-test2--skylark.int.clever.com",
		// "http://localhost:8081",
		wot2.WithLogger(wot2.NewLogger("wagoteltest", wcl.Info)),

		wot2.WithInstrumentor(tracing.InstrumentedTransport),
		// wot2.WithExporter(tracing.JaegerExporter()),
		wot2.WithExporter(tracing.OtlpGrpcExporter(ctx)),
	)
	return

}`

ChrisScotMartin: Totally open to suggestions here as I don't like the fatal log, but also don't want to complicate creating a client more than necessary.

tracing/tracing.go
```diff
@@ -0,0 +1,107 @@
...
+}
+
+func (o addrOption) apply(opts *options) {
+	opts.address = o.address
+}
+
+// OtlpGrpcExporter uses the otlptracegrpc modules and the otlptrace module to produce a new exporter at our default addr
+func OtlpGrpcExporter(ctx context.Context, opts ...Option) (sdktrace.SpanExporter, error) {
+
+	const DefaultCollectorHost = "localhost"
```
andruwm, line 43: ah sorry, meant these should be defined at the top of the file, with an explanation

andruwm (approved):
🎉

tracing/tracing.go
ChrisScotMartin: I'm curious, why? I would think it would make sense to keep them in the context of the function, but I have seen that convention of defining consts at the top before.

andruwm: If a developer is ever trying to find default values, it is much easier to first check at the top of the file instead of digging through functions

PR #448, opened 2022-10-28
accrued changes + opts in NewFromDiscovery

ChrisScotMartin:
[Jira
](https://clever.atlassian.net/browse/INFRANG-5084)

This is a fix for the problems seen in [INFRANG-5084](https://clever.atlassian.net/browse/INFRANG-5084) and should have been there from the start. 

I built all of the logic of taking options into the client's New() function but added none of that to the NewFromDiscovery() which then just ends up calling New() with no options. Since that's the real way we end up creating clients that's definitely important. 

When a new wag client was created with no options passed that meant one was created with no logger (kayvee) or tracing dependencies, which meant it defaulted to a stdout trace exporter and logger. Great for debugging, really annoying when deployed and it generates tons of logs in DD. 

This is a simple fix, NewFromDiscovery should take options and pass those options onto New()

Todo:

- [x] Run `make build`
- [x] Run `make generate`


hardcoded/hardcoded.go
```diff
@@ -1,17 +1,16 @@
-// Code generated by go-bindata. DO NOT EDIT.
```
taylor-sutton, line 1: Did the version of go-bindata change? I don't see a corresponding change in go.mod.

clients/go/gengo.go
```diff
@@ -323,15 +323,15 @@ func New(ctx context.Context, basePath string, opts ...Option) *WagClient {
 
 // NewFromDiscovery creates a client from the discovery environment variables. This method requires
 // the three env vars: SERVICE_{{.FormattedServiceName}}_HTTP_(HOST/PORT/PROTO) to be set. Otherwise it returns an error.
-func NewFromDiscovery() (*WagClient, error) {
```
taylor-sutton, line 326: This change makes me wonder if there's a different default that might be better or safer. Thoughts?

ChrisScotMartin: So I spent a while Friday experimenting with that. 

And I know it's something you mentioned here:

`
POSSIBLE ANNOYANCE: Doing things this way - where the client default is not our Clever best practice, but you have to call some library whenever you create a wag client to get it set up with tracing - is okay but not ideal, since it makes it easy to lose instrumentation by accident. Maybe there’s some way to have it all - like call some global “set default wag client options” method then whenever clients are constructed, they’ll pull from that first? Seems tricky to make work without having the client depend on e.g. particular versions of github.com/Clever/wag or the library that includes this thing, etc, but maybe.
`

The actual setup isn't "too" ugly and looks like:
```go 
	dacService, err := dacServiceClient.NewFromDiscovery(
		dacServiceClient.WithLogger(logger.NewConcreteLogger("dac-service-client")),
		dacServiceClient.WithInstrumentor(tracing.InstrumentedTransport),
		dacServiceClient.WithExporter(exporter),
	)
```

The trick is what we REALLY want to do is basically be able to pass in a client "package" to a function and then have it call that package's functions with our options. And there just isn't a way to do that in Go. 

```go 
dacService, err := someConfigPackage.AutoInstrument(dacServiceClient);
```

hardcoded/hardcoded.go
ChrisScotMartin: It did not. 

But: https://github.com/go-bindata/go-bindata/commit/6518908443d2a120be39ea957b64ac051e8f1fbd introduced that change a while ago... no idea why we are just seeing that now.

taylor-sutton: I think this may not be intended, my guess is that a `go-bindata` other than the one installed by `make install_deps` is being used.

clients/go/gengo.go
ChrisScotMartin: Noop exporter instead of stdouttrace as default exporter now. 

Also bumped stretchr/testify because Andrew did in a wagv8 PR recently, and that bumped a few other things. 

I also downgraded some of the otel stuff after testing (to the versions I was originally working with) because otel has made the decision to lock all future versions to go v1.18+ and will be quickly following new go releases and NOT considering that a breaking change as far as semver as "it's not a breaking change in the API". So I can't just get the latest of everything in the v1.x.x.

clients/go/gengo.go
```diff
@@ -62,10 +62,10 @@ import (
 
 		"go.opentelemetry.io/otel"
 		"go.opentelemetry.io/otel/propagation"
-		"go.opentelemetry.io/otel/exporters/stdout/stdouttrace"
+		"go.opentelemetry.io/otel/sdk/trace/tracetest"
 		"go.opentelemetry.io/otel/sdk/resource"
 		sdktrace "go.opentelemetry.io/otel/sdk/trace"
-		semconv "go.opentelemetry.io/otel/semconv/v1.12.0"
+		semconv "go.opentelemetry.io/otel/semconv/v1.10.0"
```
taylor-sutton, line 68: Whoa, super weird, I wonder why they decided to name their modules this way. I guess allowing different version of their specification to work in the same Go module or at least not tying the Go mod version to the spec version.

taylor-sutton (approved):

clients/go/gengo.go
ChrisScotMartin: Yeah, not a fan of that at all.

PR #452, opened 2023-01-24
Wag nightly build

ChrisScotMartin:
Todo:
[INFRANG-5176
](https://clever.atlassian.net/browse/INFRANG-5176?search_id=3e11e765-e2b6-46b1-b31a-77c1d86e66d0)

Request-Id was coming in through Envoy and that ended up getting used as our traceid. We want to keep using envoy's Request-Id and pass it along in headers but also pass along our own traceid for use in otel tracing. 

This also includes a fix for the VERSION file. 

- [x] Run `make build`
- [x] Run `make generate`


_hardcoded/tracing.go
```diff
@@ -110,37 +110,52 @@ func SetupGlobalTraceProviderAndExporterForTest() (*tracetest.InMemoryExporter,
 // Right now we only support logging IDs in the format that Datadog expects.
 func MuxServerMiddleware(serviceName string) func(http.Handler) http.Handler {
 	otlmux := otelmux.Middleware(serviceName, otelmux.WithPropagators(otel.GetTextMapPropagator()))
+	fmt.Println("Adding mux server middleware")
```
jakegut, line 113: Is this intentional?

ChrisScotMartin: Nope, removed in a prior cleanup, but this code change came from a real client that I was testing. Thanks for catching that.

jakegut (approved):
LGTM. There are a couple unhandled `err`s when you do` bag.SetMember(..)`, IDK if you want to handle those.

ChrisScotMartin:
> LGTM. There are a couple unhandled `err`s when you do `bag.SetMember(..)`, IDK if you want to handle those.

Made change, luckily it returns the original baggage on an error so no need to throw an error, but should log it somewhere. Added the error to the generated span, since this only becomes an issue if tracing is enabled, the particular span is sampled, and it for some reason has an error adding to that spans baggage.

PR #454, opened 2023-01-28
Infrang 5181

nbhatia823:
This PR:
1. Adds in the clientconfig module staring at version 9, to match wag
2. Adds ClientLogger, WithoutTracing, and WithTracing to be exported helper functions that can be used when creating wag clients.
3. Adds all tracing related code that previously existed in the generated clients.
4. Leaves the existing tracing module for now, but once we are fully migrated to clientconfig we can delete it.

- [ ] Run `make build`
- [ ] Run `make generate`


clientconfig/clientconfig.go
```diff
@@ -0,0 +1,28 @@
...
+
+func ClientLogger(wagAppName string) *logger.Logger {
+	return logger.NewConcreteLogger(fmt.Sprintf("%s-wagclient", wagAppName))
+}
+
+func WithoutTracing(wagAppName string) (*logger.Logger, *http.RoundTripper) {
+	return ClientLogger(wagAppName), &http.DefaultTransport
+}
+
+func WithTracing(ctx context.Context, wagAppName string, exporter sdktrace.SpanExporter) (*logger.Logger, *http.RoundTripper) {
```
nbhatia823, line 20: Needed to add `ctx` since it seems that is required for `DefaultInstrumentor()`

clientconfig/tracing.go
```diff
@@ -0,0 +1,163 @@
...
+	"go.opentelemetry.io/otel/exporters/otlp/otlptrace"
+	"go.opentelemetry.io/otel/exporters/otlp/otlptrace/otlptracegrpc"
+	"go.opentelemetry.io/otel/propagation"
+	"go.opentelemetry.io/otel/sdk/resource"
+	sdktrace "go.opentelemetry.io/otel/sdk/trace"
+	semconv "go.opentelemetry.io/otel/semconv/v1.12.0"
+	"go.opentelemetry.io/otel/trace"
+)
+
+var defaultSamplingProbability = 0.1
```
nbhatia823, line 22: Is this the correct value?

clientconfig/tracing.go
```diff
@@ -0,0 +1,163 @@
...
+
```
nbhatia823, line 23: The following code up until `newResource` is directly taken from `tracing/tracing.go`

clientconfig/tracing.go
```diff
@@ -0,0 +1,163 @@
...
+func ExtractSpanAndTraceID(r *http.Request) (traceID, spanID string) {
+	s := trace.SpanFromContext(r.Context())
+	if s.SpanContext().HasTraceID() {
+		return s.SpanContext().TraceID().String(), s.SpanContext().SpanID().String()
+	}
+	sc := trace.SpanContextFromContext(propagator.Extract(r.Context(), propagation.HeaderCarrier(r.Header)))
+	return sc.SpanID().String(), sc.TraceID().String()
+}
+
+// newResource returns a resource describing this application.
```
nbhatia823, line 114: The rest of this file is taken from the generated client code with two changes:
1. Replaced the appName concrete string with the variable appName
2. Updated determineSampling slightly as it seemed like the [function was incomplete](https://github.com/Clever/messaging-service/blob/master/gen-go/client/client.go#L197-L214) before

clientconfig/tracing.go
```diff
@@ -0,0 +1,163 @@
...
+	}
+	return spanExporter, nil
+
+}
+
+// DefaultInstrumentor returns the transport to use in client requests.
+// It takes in a transport to wrap, e.g. http.DefaultTransport, and the request
+// context value to pull the span name out from.
+// 99% sure this is wrapping a wrapped thing and totally redundant. Fix later.
+func DefaultInstrumentor(spanNameCtxValue interface{}, baseTransport http.RoundTripper, tp sdktrace.TracerProvider) http.RoundTripper {
```
nbhatia823, line 79: does `spanNameCtxValue` need to be an interface instead of `context.Context`?

nbhatia823: Same comment for line 85

clientconfig/tracing.go
```diff
@@ -0,0 +1,163 @@
...
+	"time"
+
+	"go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp"
+	"go.opentelemetry.io/otel"
+	"go.opentelemetry.io/otel/exporters/otlp/otlptrace"
+	"go.opentelemetry.io/otel/exporters/otlp/otlptrace/otlptracegrpc"
+	"go.opentelemetry.io/otel/propagation"
+	"go.opentelemetry.io/otel/sdk/resource"
+	sdktrace "go.opentelemetry.io/otel/sdk/trace"
+	semconv "go.opentelemetry.io/otel/semconv/v1.12.0"
```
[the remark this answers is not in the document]
nbhatia823: Gotcha I can keep it at 1.10, was an unintentional upgrade

clientconfig/tracing.go
nbhatia823: Nice, I agree that removing it is the right call

clientconfig/tracing.go
nbhatia823: Love that we can drop `ctx` then from `WithTracing`! 

I'm going to push a commit that does what I think you suggested, but I'm not sure so feel free to correct it

clientconfig/clientconfig.go
nbhatia823: Discussed [here](https://github.com/Clever/wag/pull/454#discussion_r1091055522) that we don't need `ctx` anymore, which is nice

ChrisScotMartin (approved):
SWEET!

PR #455, opened 2023-02-02
Update wag go client creation API

nbhatia823:
This PR updates the wag client creation API to require a logger and transport, as discussed in this [spec](https://docs.google.com/document/d/19eNjVWxx7xAJ1RWzfdli4CN0vaMuwoEXX-3fNlHAnto/edit). 

Look at this[ branch of messaging-service ](https://github.com/Clever/messaging-service/compare/test-new-wag-client?expand=1#diff-2873f79a86c0d8b3335cd7731b0ecf7dd4301eb19a82ef7a1cba7589b5252261R81)to see what it would look like to create a client in this new world.

This[ file is an example](https://github.com/Clever/messaging-service/blob/29d5194f25e16f4403987ca23904547addf6e91e/gen-go/client/client.go) of what a new generated client will look like.

- [ ] Run `make build`
- [ ] Run `make generate`
- [ ] Update the current version in the `/VERSION` file. WILL DO BEFORE MERGING

### DIFF
--- a/clients/go/gengo.go
+++ b/clients/go/gengo.go
@@ -46,7 +46,6 @@ import (
 		"strconv"
 		"time"
 		"fmt"
-		"os"
 		"io/ioutil"
 		"crypto/md5"
 
@@ -57,13 +56,6 @@ import (
 
 
 		"github.com/afex/hystrix-go/hystrix"
-
-		"go.opentelemetry.io/otel"
-		"go.opentelemetry.io/otel/propagation"
-		"go.opentelemetry.io/otel/sdk/trace/tracetest"
-		"go.opentelemetry.io/otel/sdk/resource"
-		sdktrace "go.opentelemetry.io/otel/sdk/trace"
-		semconv "go.opentelemetry.io/otel/semconv/v1.10.0"
 		
 )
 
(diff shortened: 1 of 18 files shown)


clients/go/gengo.go
```diff
@@ -93,241 +85,63 @@ type WagClient struct {
 
 var _ Client = (*WagClient)(nil)
 
+// New creates a new client. The base path, logger, and http transport are configurable.
+// The logger provided should be specifically created for this wag client. If tracing is required,
+// provide an instrumented transport using the wag clientconfig module. If no tracing is required, pass nil to use
+// the default transport.
+func New(basePath string, logger wcl.WagClientLogger, transport *http.RoundTripper) *WagClient {
```
ChrisScotMartin, line 92: Total nit: I feel a little weird about basePath as a name because we over use it so much in wag talking about dir base paths etc.  serviceName seems more descriptive/relevant to me.

ChrisScotMartin (approved):
So happy about this! 🕺 🥳 🎉

clients/go/gengo.go
nbhatia823: I'm a little hesitant to change the param name since its been around for a while and people are pretty familiar with it. Also I dont think that serviceName is correct here - for ex. [here](https://github.com/Clever/app-view-service/blob/master/main.go#L177-L181). basePath is meant to be a path in this [format](https://github.com/Clever/discovery-go/blob/master/discovery.go#L42)

PR #460, opened 2023-07-27
Swad/infrang 4971

swadkumar:
**JIRA:** https://clever.atlassian.net/browse/INFRANG-4971

**Overview:**

_What does this change do?_
>This change attempts to fix the error caused by unused import of aws by using a local function to 
   return pointer to a string

_Why is this change needed?_
>Avoid unused import bug in DynamoDB tests generated using wag template

_How was this change tested?_
> Tested by generating samples from template in WAG and checking syntax

**Checklist:**

- [x] Run `make build`
- [x] Run `make generate`


VERSION
```diff
@@ -1,5 +1,7 @@
 v9.1.0
 
+v9.1.0 Fixes for unused aws import in DynamoDB tests
+
```
ChrisScotMartin, line 4: So we were already on v9.1.0. This file should only be updated when/if you bump the version. 

Bumping to v9.1.1 is perfectly fine.

ChrisScotMartin (changes requested):
Looks good, just the version update mention. And then be sure to `make generate` after that. 

What ended up being the cause of your `make generate` issues before?

VERSION
swadkumar: Got it, will update the version. 

make generate wasn't working due to me using go1.20. This works with go1.16.

andruwm (approved):

PR #462, opened 2023-08-31
Swad infrang 4840

swadkumar:
**Overview:**
TKT: https://clever.atlassian.net/browse/INFRANG-4840/

**Issue in messaging-service:**
https://github.com/Clever/messaging-service/commit/1520c9d4c1cd1e6a442d845e1e53471ed590f5f3#diff-753a7037d67b737e8d239e4e9539e987c1015bf01a36621e4f17196269f6f230R169

**Note:**
The workaround in messaging-service needs to be removed as well after this fix:
https://github.com/Clever/messaging-service/blob/master/Makefile#L43-L45

Todo:

- [X] Run `make build`
- [X] Run `make generate`

### DIFF
--- a/VERSION
+++ b/VERSION
@@ -108,5 +108,3 @@ example go.mod file under samples/.
 - v3.15.0: JS client enable compression
 - v3.14.3: Remove unused code.
 - v3.14.2: Stricter span checks to prevent tracing errors.
-- v3.14.1: Send client version as a header bugfix.
-- v3.14.0: Send client version as a header.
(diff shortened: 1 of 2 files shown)


VERSION
```diff
@@ -1,4 +1,6 @@
-v9.1.2
+v9.1.3
+
+v9.1.3 Fix for generated code supporting filtering on an enum
```
ChrisScotMartin, line 3: It's generally best practice when adding a new line here to remove the bottom one so our history here doesn't get too large.

ChrisScotMartin: But it looks like we haven't been doing that here?

swadkumar: resolved

tnsardesai (approved):

PR #471, opened 2024-01-17
Http2! (it's actually h2c)

tnsardesai:
#Jira 
https://clever.atlassian.net/browse/INFRANG-5794

# Overview
Took some examples from https://github.com/aws/aws-app-mesh-examples/blob/main/walkthroughs/howto-http2/color_server/main.go#L85-L89. 

We use envoy for TLS handshake for internal services and for external services the TLS is terminated by the ALB so our wag servers need to use h2c (which is http/2 - TLS)

# Todo
- [x] Run `make build`
- [x] Run `make generate`

# Testing

Deployed sso-il-token-service with these changes in tanmay-dev. And with the help of https://github.com/Clever/catapult/pull/2155 tested that http2 works! 

I am also going to run a staging load test!

#


jakegut (approved):

tnsardesai:
We are not pursuing this update right now so I am gonna close the PR :)

PR #474, opened 2024-03-04
fix dynamoodb codegen for datetime partition keys for GSIs

dxu429:
# Jira
https://clever.atlassian.net/browse/DD-5016

# Overview
~1. Fix import for dynamodb - should only be imported in dynamodb.go when there are transactions~
    - created a new PR for the above: https://github.com/Clever/wag/pull/475.
2. Fix dynamodb codegen for `datetime` `HASH` keys on GSIs

Currently code gen for date-time `HASH` keys for GSI's doesn't properly convert `strfmt.Datetime` to `strings` in the generated getBy{GSI} functions.

# Example
here `input.endDate` has type `strfmt.Date`, which should be converted to a string 
## Before
```
 <<<<<<< type mismatch>>>>>>>>
if input.EndDate == "" {
 <<<<<<< type mismatch>>>>>>>>
	return fmt.Errorf("Hash key input.EndDate cannot be empty")
}
queryInput := &dynamodb.QueryInput{
	TableName: aws.String(t.name()),
	IndexName: aws.String("endDate"),
	ExpressionAttributeNames: map[string]*string{
		"#ENDDATE": aws.String("endDate"),
	},
	ExpressionAttributeValues: map[string]*dynamodb.AttributeValue{
		":endDate": &dynamodb.AttributeValue{
<<<<<<< type mismatch>>>>>>>>
			S: aws.String(input.EndDate),
<<<<<<< type mismatch>>>>>>>>
		},
	},
	ScanIndexForward: aws.Bool(!input.Descending),
	ConsistentRead:   aws.Bool(false),
}
```

### After
```
if toDynamoTimeString(input.EndDate) == "" {
	return fmt.Errorf("Hash key input.EndDate cannot be empty")
}
queryInput := &dynamodb.QueryInput{
	TableName: aws.String(t.name()),
	IndexName: aws.String("endDate"),
	ExpressionAttributeNames: map[string]*string{
		"#ENDDATE": aws.String("endDate"),
	},
	ExpressionAttributeValues: map[string]*dynamodb.AttributeValue{
		":endDate": &dynamodb.AttributeValue{
			S: aws.String(toDynamoTimeString(input.EndDate)),
		},
	},
	ScanIndexForward: aws.Bool(!input.Descending),
	ConsistentRead:   aws.Bool(false),
}
```

- [x] Run `make build`
- [x] Run `make generate`

# Testing
1. Created a sample schema to test `ThingWithDatetimeGSI`
2. Tested by building & using built executable to `make generate` on a service I'm using this on. https://github.com/Clever/order-service/compare/master...subscriptions-model
Screenshot 2024-03-04 at 12 10 21 PM
Screenshot 2024-03-04 at 12 11 49 PM


tnsardesai (commented):
https://github.com/Clever/catapult/blob/master/swagger.yml#L2088-L2110. I haven't taken a look proper look at this PR but I know we are doing something similar in catapult. What is the difference here? Looks like in catapult - id is the `HASH` and time is `RANGE` and here it is in reverse

dxu429:
@tnsardesai I'm using a datetime for the `HASH` key

andruwm (approved):

PR #483, opened 2024-04-15
[INFRANG-5930] Remove default table names

andruwm:
# Jira
https://clever.atlassian.net/browse/INFRANG-5930

# About

The default table name provided is not Clever's standard table naming convention. Because of the default behavior, this led to cryptic IAM errors at run time when an engineer forgot to specify the table name when adding a new table. These errors had a table name that was often nearly identical, aside from capitalization, to the real table name. Getting rid of the default has no real effect besides making this step easier to perform and debug.

Todo:

- [x] Run `make build`
- [x] Run `make generate`

# Testing
Verified using samples generated code


tnsardesai (approved):
LGTM

PR #486, opened 2024-04-20
[INFRANG-5930] fix unused fmt import

andruwm:
## Jira
https://clever.atlassian.net/browse/INFRANG-5930

## About
fmt is referenced 40 times, all of which are conditional. Rather than trying to account for many cases in which fmt can be used, we can add an unconditional usage to circumvent the problem.

## Testing
Ran make test

## Checklist
- [x] Run `make build`
- [x] Run `make generate`


swadkumar (approved):

taylor-sutton:
> we can add an unconditional usage to circumvent the problem.

Your fix seems good - enhancing an error message. Another option could have been to use an underscore thing like this https://github.com/Clever/wag/blob/42d4590c996d1f7c392f66bb1e780cb940afbc65/samples/gen-go-db/server/db/dynamodb/thingwithunderscores.go#L16

PR #490, opened 2024-07-24
fix warn to warning for 4xx logs

jakegut:
## Jira

https://clever.atlassian.net/browse/INFRANG-6286

## Testing

* Tested on broken branch mentioned in Slack

## Checklist
- [ ] Run `make build`
- [ ] Run `make generate`
- [ ] Update the current version in the `/VERSION` file. First line should be JUST the version e.g. `v10.13.1`. Then a blank line. Then release notes.

### DIFF
--- a/VERSION
+++ b/VERSION
@@ -1,4 +1,6 @@
-v9.3.2
+v9.3.3
+
+v9.3.3 Fix typo for generating warning logs
 
 v9.3.2 Fix string enums as GSIs for dynamodb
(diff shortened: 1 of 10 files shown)


taylor-sutton (approved):
As a follow-up we should understand why this wasn't caught by CI.

jakegut:
I took a look and it seems like `make test` doesn't build the samples which would've caught this

PR #493, opened 2024-10-17
Support date formatted strings in DynamoDB codegen for GSIs

dxu429:
# Jira

https://clever.atlassian.net/browse/DD-5757

# Overview
**Review by commit**

Adds support for `strfmt.Date` typed fields in dynamodb keys, currently the codegen does not support it. 

# Testing
plz soc2 why won't you recognize me
## Before
Created some test cases in `db.yml` then ran make generate without the change
Screenshot 2024-10-17 at 1 34 12 PM
Screenshot 2024-10-17 at 1 34 30 PM

## After
After updating the templates and `gendb` files
Screenshot 2024-10-17 at 1 54 46 PM
Screenshot 2024-10-17 at 1 55 11 PM

# Checklist
- [x] Run `make build`
- [x] Run `make generate`
- [x] Update the current version in the `/VERSION` file. First line should be JUST the version e.g. `v10.13.1`. Then a blank line. Then release notes.

PR #494, opened 2024-10-18
Fix transact bug

dxu429:
# Jira

https://clever.atlassian.net/browse/DD-5762

# Overview
While testing the wag build from https://github.com/Clever/wag/pull/493 in a service where I was attempting to add a GSI using a `date` formatted string as hash key, I found wag was auto generating transaction functions multiple times.
Turns out the template to autogen transact functions are in the wrong range loop; currently they are in the GSI {{ range }} block but transact funcs should operate outside/independently of GSIs
**review by commit**

# Testing
- Tested against my own use case 
- Created a test case in db.yml for a before and after:
**Before**
Screenshot 2024-10-18 at 1 29 18 PM
**After**
Screenshot 2024-10-18 at 1 32 30 PM

# Checklist
- [x] Run `make build`
- [x] Run `make generate`


VERSION
```diff
@@ -1,4 +1,6 @@
-v9.4.0
+v9.4.1
+
+v9.4.1 Fix bug in transaction codegen where transaction funcs where generated multiple times
```
dxu429, line 3: any opinions on roll this fix with v9.4.0 so only 1 release needs to be made?

andruwm (approved):
I think its fine to roll them together, just make sure you run make generate again after you merge this PR into your other one!

PR #501, opened 2024-12-17
baggage consistently a map now

ChrisScotMartin:
A few thing were done here:
- wag-js baggage is now a map of string to string or number. At one point it was an object and there were some artifacts of that left over that needed changed/removed.

- Server tracing logic no longer sets otel generated traceids as `clever-request-id` these are added to logger context and saved in datadog format if tracing is enabled and otherwise we don't need it. 

- If a clever-request-id isn't set we generate a uuid and set that as clever-request-id. 

- Cleaned up some logic for "merging baggage headers" in wag-js clients. This never needs to be a thing, as if we want to set baggage in wag-js we should just do it by passing baggage, we don't even have a way of passing headers. That logic was dumb. And I can say that because I wrote it. 

## Jira

https://clever.atlassian.net/browse/INFRANG-6614

## Testing
Generated district-config-service using this version of wag, imported that client into app-views-service, generated that with this version, ditto for il-token-service, and then finally imported the generated wag-js client of il-token-service into instant-login-api. 
Then tested propagation ( instant-login-api -> sso-il-token-service -> sso-app-view-service -> sso-district-config-service
) by disabling rollups in each of these services and deploying them into my environment. 

Started `il-token-service` locally in the `csm` env and  ran 

```bash
curl -X GET http://localhost:6001/v2.1/me \
     -H "clever-request-id: 123456" \
     -H "baggage:user-id=67890,session-id=abcerw3e3534*(#$,clever-request-id=999" \
     -H "Authorization: Bearer clever_user_token_string"
```
and saw the expected propagation:
image

Also tested with some fuzzing on the baggage header to catch any more possible issues. (Note: It is possible to generate an invalid baggage header by doing this, this would then cause the baggage propagation to fail as otel does no sanitation of the header. This is fine as baggage will always be set and controlled by us. And we do sanitation of the header at the top of the request chain. This testing was to make sure we fail gracefully, preserve valid baggage key/pairs and don't end up with any nil pointer dereferences.)

Tested with and without adding a clever-request-id to baggage. If it didn't exist it should create a uuid, if it does it should pass that id. 
Tested with a `clever-request-id` header that we used to set and made sure that doesn't override anything anymore. 
Disabled tracing in half of the wag services, and enabled it in the other half of the chain to test edge cases around tracing. 
Continued by testing without passing `clever-request-id` and seeing it generate a uuid that was then propagated. 

## Checklist
- [x] Run `make build`
- [x] Run `make generate`


_hardcoded/tracing.go
```diff
@@ -121,19 +123,19 @@ func MuxServerMiddleware(serviceName string) func(http.Handler) http.Handler {
...
-			crid = s.SpanContext().TraceID().String()
 			bags := baggage.FromContext(ctx)
-			// Add clever-request-id to baggage
-			member, err := baggage.NewMember("clever-request-id", crid)
-			if err != nil {
-				logger.FromContext(ctx).WarnD("error-creating-baggage", logger.M{"error": err.Error()})
-			} else {
-				bags, err = bags.SetMember(member)
+
+			if bags.Member("clever-request-id").String() == "=" { // if clever-request-id is not set
```
ChrisScotMartin, line 129: Otel is dumb and they don't have a way of checking if a baggage key exists. Their get returns an "empty member" if it doesn't exist which isn't nil or "" so you can't compare to something simple like that. If you attempt to get the value of something that doesn't exist though you get a nil pointer dereference. Which makes no sense to me. 

So to check if something doesn't exist the safe/low effort way of doing it is call their String() function on it which returns `<key>=<value>` and if this member doesn't exist that ends up just being `=`

You can shoot me now.

_hardcoded/tracing.go
```diff
@@ -145,15 +147,16 @@ func MuxServerMiddleware(serviceName string) func(http.Handler) http.Handler {
...
 			// Encode the trace/span ids in the DD format
 			if sc := s.SpanContext(); sc.HasTraceID() {
+
+				// Log if sampled
+				if s.SpanContext().IsSampled() {
+					logger.FromContext(ctx).AddContext("sampled", "true")
+				} else {
+					logger.FromContext(ctx).AddContext("sampled", "false")
+				}
+
```
ChrisScotMartin, line 159: No need to log if it's sampled or not if we aren't actually exporting the "sampled" trace.

VERSION
```diff
@@ -1,4 +1,6 @@
-v9.5.1
+v9.5.2
+
+v9.5.2 Bug fix in baggage options for JS client, cleanup tracing.go a bit
```
andruwm, line 3: I dont like that our SEMVER format restricts us to one line descriptions of change, but this is pretty vague.

andruwm (approved):
One version description nit but lgtm

VERSION
taylor-sutton: It doesn't, the release notes end up being the entirety of this file https://github.com/Clever/wag/releases

ChrisScotMartin: 👍  Updated with a slightly more verbose message.

andruwm: > It doesn't, the release notes end up being the entirety of this file https://github.com/Clever/wag/releases

Ah maybe I am thinking of ark

PR #503, opened 2025-03-04
Infrang 6864

ChrisScotMartin:
## Jira
https://clever.atlassian.net/browse/INFRANG-6864
https://clever.atlassian.net/browse/INFRANG-6851

## Description
Go tests have not been running for an extended period of time ( > 1 year) on our generated samples.

Go tests were fixed so that they were compilable and matched the function signatures expected. Then some minor bug fixes were added to tests. Unfortunately MANY tests are failing and presumably would have been failing for an extended time. So for now I've disabled tests in CI. 

This WILL be a temporary measure. The next step will be to add t.Skip() to each test that fails and a ``FAILING_TEST`` comment to tag each of these, then re-enable CI tests. 

After that we can tackle fixing tests through future tickets.

For now `make test` should be run manually before and after large changes and the diff compared. This is cumbersome and not a good state to leave us in even mid term. 

## Testing
`make release` then extracted binary and moved it over to `user-mgmt-events-service` disabled wag version checking/updating in `user-mgmt-events-service`Makefile then used new binary to rebuild generated files with `make generate`

Then committed those changes to `testing-new-client` branch. 

Moved over to `library-customization-service` and ran `go get github.com/Clever/user-mgmt-events-service/gen-go/client@testing-new-client` and added a quick loop to call out to `user-mgmt-events-service` using the newly generated clients and models. 

```go
	// health check
	go func() {
		for {
			time.Sleep(5 * time.Second)
			log.Default().Println("Health check:")
			spew.Dump(umesClient.HealthCheck(ctx))
			result, err := umesClient.GetAllEventActionConfigs(ctx, "5745c0ace24a810100000040")
			if err != nil {
				log.Default().Println("Error getting event action configs:")
				spew.Dump(err)
			} else {
				log.Default().Println("Event action configs:")
				spew.Dump(result)
			}
		}
	}()
```

TODO

## Checklist
- [x] Run `make build`
- [x] Run `make generate`


samples/test/server_test.go
```diff
@@ -399,15 +411,15 @@ func TestTimeout(t *testing.T) {
 	start := time.Now()
 	_, err = c.GetBooks(ctx, &models.GetBooksInput{})
 	assert.Error(t, err)
-	assert.IsType(t, context.DeadlineExceeded, err)
+	assert.Contains(t, err.Error(), "context deadline exceeded")
```
ChrisScotMartin, line 414: We return a *url.Error with text that includes "context deadline exceeded" now.

samples/test/tracing_test.go
ChrisScotMartin, line 1: Fully removed this as many things about this test were broken and since we aren't actively using tracing it's not worth fixing right now.

samples/Makefile
```diff
@@ -11,40 +11,74 @@ generate:
 	@echo  $(TEST_PKGS) $(TEST_DB_PKG)
 	
 	../bin/wag -with-tests -file ./swagger.yml -output-path ./gen-go-basic -js-path ./gen-js
-	cd ./gen-js && jsdoc2md index.js types.js > ./README.md
+	cd ./gen-js && jsdoc2md index.js types.js > ./README.md    
+	cd ./gen-go-basic/client && go mod tidy
+	cd ./gen-go-basic/models && go mod tidy
```
ChrisScotMartin, line 16: We need to properly fill out submodule dependencies and we weren't doing that which was one cause of these tests being unable to run because client/models were not compilable.

samples/Makefile
```diff
@@ -11,40 +11,74 @@ generate:
...
+
 	../bin/wag -with-tests -file ./db.yml -output-path ./gen-go-db-custom-path -js-path ./gen-js-db-custom-path -dynamo-path db
 	cd ./gen-js-db-custom-path && jsdoc2md index.js types.js > ./README.md
+	cd ./gen-go-db-custom-path/client && go mod tidy
+	cd ./gen-go-db-custom-path/models && go mod tidy
+	cd ./gen-go-db-custom-path/ && go mod tidy
 	go generate -mod=mod ./gen-go-db-custom-path...
 
+	go install -mod=mod golang.org/x/tools/cmd/goimports@v0.24.0
+	goimports -w .
```
ChrisScotMartin, line 75: I use goimports here to clean out unused dependencies from our go files. This is only an issue because of our db-only builds that have end up including the models but then not using them in certain files. This is not a pattern that we see actually happening at Clever.

samples/Makefile
```diff
@@ -11,40 +11,74 @@ generate:
...
+
  $(TEST_PKGS) $(TEST_DB_PKG): golang-test-all-strict-deps
-	$(call echo golang-test-all,$@)
+	$(call golang-test-all,$@)
 
 install_deps:
+	go mod tidy
```
ChrisScotMartin, line 81: We should always tidy as part of vendor post go 1.16.

VERSION
```diff
@@ -1,4 +1,7 @@
-v9.5.3
+v9.6.0
+remove hystrix-go dependency from wag go clients, fix some tests.
```
tnsardesai, line 2: I would have preferred 2 PRs, 1 for clean up and 1 for removal of hystrix-go

clients/go/gengo.go
```diff
@@ -99,32 +94,19 @@ func New(basePath string, logger wcl.WagClientLogger, transport *http.RoundTripp
...
-		// one circuit for each service + url pair
-		circuitName: fmt.Sprintf("{{.ServiceName}}-%%s", shortHash(basePath)),
-		logger: logger,
-	}
-	circuit.init()
-
 	client := &WagClient{
 		basePath: basePath,
-		requestDoer: circuit,
+		requestDoer: &base,
```
tnsardesai, line 102: Should this be `retryDoer{d: base, retryPolicy: SingleRetryPolicy{}}`? We want to retry at least 1 time.

samples/Makefile
tnsardesai, line 16: Can you use a function here. For most of the cases it looks like this. 
```
define gen-tidy
@cd $(1) && jsdoc2md index.js types.js > ./README.md
@cd $(2)/client && go mod tidy
@cd $(2)/models && go mod tidy
enddef
```

ChrisScotMartin: Most of them are, so I suppose I could use a function for most.

clients/go/gengo.go
ChrisScotMartin: Nice catch, annoyingly I did this right the first time around and missed it this time 😞

samples/Makefile
ChrisScotMartin: Hmm, good idea, if I'm going that route might as well clean this up even more.

tnsardesai (approved):

PR #504, opened 2025-03-05
[INFRANG-6876] Upgrade to go 1.24

andruwm:
# JIRA
https://clever.atlassian.net/browse/INFRANG-6876

# About
This PR upgrades this repo to Go version 1.24. See the [release notes](https://tip.golang.org/doc/go1.24) for full details.

It performs the following changes:
1. Upgrade the module to 1.24.
2. Upgrade the golang.mk file to the latest version.
3. Replaces any `tools.go` file with the new go.mod tools directive.
4. Updates any circle CI build images to 1.24.
5. Updates the go version check in the makefile.
6. Updates any debian docker images to a version with a glibc compatible with the CI build image.

## WAG Specific changes
There were a lot of `go vet` checks to fix. I had to add a new print function in order to have the correct function to call. This resulted in a bunch of `%%` characters in generated code. In `printf` `%%` is interpreted as a literal `%`. Since we are now using the correct method to print, it is just printed as is. 

#### New Go Vet Check
Go vet in 1.24 introduces a new check which catches non-constant format strings in calls to printf functions. This has a tendency to find bugs in existing code, however the fix is pretty strait forward should your CI begin to fail. You can see more details in the official issue https://github.com/golang/go/issues/60529. If your CI fails because of this, you may add an extra commit resolving the issue, then merge.

# Testing
Go version upgrades have historically been extremely stable. The only exception has been incompatible glibc versions which have been tested for in workers before merging. All standard CI testing is also performed before merging.

Since I bumped all of the submodules to 1.24, I tested out compatibility in Catapult by using my local wag to generate there and ensure it compiles, which it does! This should simplify the migration process.

# Problems Upgrading?
Checkout this [google doc](https://docs.google.com/document/d/1ctg3eT8zkKHXqsf7CCUBB77ERmEVMsyZoYAIxXm19hs/edit?usp=sharing) for knowledge sharing any problems you encounter during upgrades!


go.mod
```diff
@@ -65,3 +65,8 @@ replace gopkg.in/yaml.v2 => gopkg.in/yaml.v2 v2.0.0
 // Newer versions of swag use a newer version of yaml which fix a "bug" where the yaml spec was not fully
 // addhered  to. Unfortunately this was a breaking change in yaml parsing, so pinning swag for now.
 replace github.com/go-openapi/swag => github.com/go-openapi/swag v0.21.1
+
+tool (
+	github.com/golang/mock/mockgen
+	github.com/kevinburke/go-bindata/go-bindata
+)
```
ChrisScotMartin, line 72: So excited about this!

swadkumar (approved):
awesome!

PR #505, opened 2025-03-07
Infrang 6880

ChrisScotMartin:
## Jira

https://clever.atlassian.net/browse/INFRANG-6880

## Description
With wag tests now running they will fail CI so they’ve been temporarily disabled. We don’t have good visibility into just how many tests are failing because when one test file fails it doesn’t run the next.

Marked individual failing tests with FAILED_TESTS and skipped them. 

Still had to disable all db tests due to some generated tests failing that were not easy to skip. 

All remaining failing tests are in db tests so re-enabling CI testing 

Added a comment and reference to the ticket about DB tests not running so it's not a surprise. 

## Testing

Tests are what I was fixing here so... `make test` :-) 

## Checklist
- [ ] Run `make build`
- [ ] Run `make generate`
- [ ] Update the current version in the `/VERSION` file. First line should be JUST the version e.g. `v10.13.1`. Then a blank line. Then release notes.


taylor-sutton (approved):
Nice work! I found it very easy from here to re-enable DB tests and investigate further, which is a good state to be in.

Having done that, I think you can improve the situation as far as explicitly marking which tests are failing with this patch:

```
modified   server/gendb/tests.go.tmpl
@@ -65,7 +65,8 @@ func RunDBTests(t *testing.T, dbFactory func() db.Interface) {
   {{- end }}
   {{- range $transactWithSchemaName := $xdbConfig.EnableTransactions }}
   {{- $transactWithModelName := pascalize $transactWithSchemaName }}
-  t.Run("TransactSave{{ $modelName }}And{{ $transactWithModelName }}", TransactSave{{ $modelName }}And{{ $transactWithModelName }}(dbFactory(), t))
+  /* FAILING TEST */
+  // t.Run("TransactSave{{ $modelName }}And{{ $transactWithModelName }}", TransactSave{{ $modelName }}And{{ $transactWithModelName }}(dbFactory(), t))
   {{- end }}
   {{- end }}
         {{- end }}
```

And then at a glance, it seems like there are only two tests remaining that are failing but not explicitly marked, which seem like they have to do with datetime.

Is there any point in releasing a new wag version (say, a patch version with this change?

ChrisScotMartin:
> /* FAILING TEST */

Are you thinking I should make this change as part of this PR? It was hard to tell when to stop so I kept myself to a pretty strict time box. 

And maybe? I debated doing that and chose not to as no application code was changed. Everything modified was in the build/test pipeline.

PR #507, opened 2025-04-09
[INFRANG-6691] Upgrade otel deps to stable versions where possible, and use latest kayvee which eliminated otel dep

andruwm:
## Jira
https://clever.atlassian.net/browse/INFRANG-6991

## About
This should help a lot with otel dep hell moving forwards 

## Testing
These are dependency upgrades only so there is nothing made by clever to test.

## Checklist
- [x] Run `make build`
- [x] Run `make generate`


andruwm:
Turns out the newer version of otelhttp requires go 1.22, so this will have to wait until after the migration is complete

swadkumar (approved):

PR #516, opened 2025-08-11
DIOC-1494: Adds support for configurable js client timeouts via x-cli…

trupti-clever:
…ent-timeout extension

## Jira
[DIOC-1494](https://clever.atlassian.net/browse/DIOC-1494)

## Testing: 
make test

## Checklist
- [x] Run `make build`
- [x] Run `make generate`

[DIOC-1494]: https://clever.atlassian.net/browse/DIOC-1494?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/VERSION
+++ b/VERSION
@@ -1,3 +1,6 @@
+v9.9.6
+Add support for configurable client timeouts via x-client-timeout Swagger extension (JavaScript clients only)
+
 v9.9.5
 Fix dynamodb codegen for paging in scan functions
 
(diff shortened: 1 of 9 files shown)


andruwm:
Since timeout is already configurable via client settings, can you please add some explanation and/or justification to the ticket and PR for why we also need this to be configurable via the swagger file?

trupti-clever:
> Since timeout is already configurable via client settings, can you please add some explanation and/or justification to the ticket and PR for why we also need this to be configurable via the swagger file? Especially because this only works on JS clients of Go services.

i wanted it so that we could configure the timeout for all methods of client. But I spoke to the team, we don't infact need this.

